PR #23, opened 2019-10-17
Fix issue with using UUID primary keys in complex return types

cking-vonix:
When using a UUID as a primary key with PostgreSQL the grouping was
defaulting to the row which caused incorrect results to be returned.


cking-vonix:
Creating new pull request onto develop

PR #47, opened 2020-08-18
Schema rename support

### LINKED ISSUE #46: Schema rename support
Add support for renaming the schema for generated code. Currently there is no way to use the generated code for identical database structure using two different schema names. 

Use cases for this includes multitenant systems and/or using different schema name in test database than in development in order to avoid date-destroying accidents in production database. The issues is even more serious in MySQL, as the schema name is in fact the name of the database.

Furthermore, it would be nice to have support for using the default schema (i.e. leave out the schema name part in the SQL queries) for the above use cases.

Support can be build just like the `AS` -function is supported in the generated code. This allows the same generated code to access tables in multiple schemas even simultaneously.

joonash:
Resolves #46

- Support for renaming table schemas
 * Table support for renaming schema
 * Empty schema name is left out (using default schema for the
   database connection)
 * Generated code support for obtaining a version of the table with
   renamed schema, similarly as the `AS` function works
 * Unit tests for setting and clearing the schema name

### DIFF
--- a/generator/internal/template/templates.go
+++ b/generator/internal/template/templates.go
@@ -41,6 +41,15 @@ type {{.GoStructName}} struct {
 func (a *{{.GoStructName}}) AS(alias string) {{.GoStructName}} {
 	aliasTable := new{{.GoStructName}}()
 	aliasTable.Table.AS(alias)
+	aliasTable.Table.Schema(a.Table.SchemaName())
+	return aliasTable
+}
+
+// Schema creates new {{.GoStructName}} with assigned schema name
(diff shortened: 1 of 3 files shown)


go-jet:
Looks good.  
You'll probably have to update generator tests, because you've changed generator template. Take a look at [postgres](https://github.com/go-jet/jet/blob/master/tests/postgres/generator_test.go) and [mysql](https://github.com/go-jet/jet/blob/master/tests/mysql/generator_test.go) tests.

PR #50, opened 2020-08-26
Add LATERAL support for postgres

### LINKED ISSUE #49: Add support for LATERAL for Postgres
Add support for the LATERAL operator.

OrlandoCo:
Resolves #49

* Add supports for pg:
  * LATERAL
  * INNER JOIN LATERAL
  * LEFT JOIN LATERAL
  * RIGHT JOIN LATERAL
  * FULL JOIN LATERAL
  * CROSS JOIN LATERAL

 All receives a SelectTable interface

### DIFF
--- a/internal/jet/lateral_table.go
+++ b/internal/jet/lateral_table.go
@@ -1,0 +1,41 @@
+package jet
+
+// LateralTable is interface for Lateral sub-queries
+type LateralTable interface {
+	Serializer
+	Alias() string
+	AllColumns() ProjectionList
+}
+
+type lateralTableImpl struct {
+	selectStmt SerializerStatement
+	alias      string
+}
+
+// NewLateralTable func
+func NewLateralTable(selectStmt SerializerStatement, alias string) LateralTable {
+	lateralTable := &lateralTableImpl{selectStmt: selectStmt, alias: alias}
+	return lateralTable
+}
+
+func (s lateralTableImpl) Alias() string {
+	return s.alias
+}
+
+func (s lateralTableImpl) AllColumns() ProjectionList {
+	statementWithProjections, ok := s.selectStmt.(HasProjections)
+	if !ok {
+		return ProjectionList{}
+	}
+
+	projectionList := statementWithProjections.projections().fromImpl(s)
+	return projectionList.(ProjectionList)
+}
+
+func (s lateralTableImpl) serialize(statement StatementType, out *SQLBuilder, options ...SerializeOption) {
+	out.WriteString("LATERAL")
+	s.selectStmt.serialize(statement, out)
+
+	out.WriteString("AS")
+	out.WriteIdentifier(s.alias)
+}
--- a/internal/jet/table.go
+++ b/internal/jet/table.go
@@ -92,11 +92,6 @@ const (
 	RightJoin
 	FullJoin
 	CrossJoin
-	InnerJoinLateral
-	LeftJoinLateral
-	RightJoinLateral
-	FullJoinLateral
-	CrossJoinLateral
 )
 
 // Join expressions are pseudo readable tables.
@@ -174,16 +169,6 @@ func (t *joinTableImpl) serialize(statement StatementType, out *SQLBuilder, opti
 		out.WriteString("FULL JOIN")
 	case CrossJoin:
 		out.WriteString("CROSS JOIN")
-	case InnerJoinLateral:
-		out.WriteString("INNER JOIN LATERAL")
-	case LeftJoinLateral:
-		out.WriteString("LEFT JOIN LATERAL")
-	case RightJoinLateral:
-		out.WriteString("RIGHT JOIN LATERAL")
-	case FullJoinLateral:
-		out.WriteString("FULL JOIN LATERAL")
-	case CrossJoinLateral:
-		out.WriteString("CROSS JOIN LATERAL")
 	}
 
 	if utils.IsNil(t.rhs) {
@@ -192,12 +177,7 @@ func (t *joinTableImpl) serialize(statement StatementType, out *SQLBuilder, opti
 
 	t.rhs.serialize(statement, out)
 
-	if t.joinType >= InnerJoinLateral && t.joinType != CrossJoinLateral {
-		out.WriteString("ON true")
-		return
-	}
-
-	if t.onCondition == nil && t.joinType != CrossJoin && t.joinType < InnerJoinLateral {
+	if t.onCondition == nil && t.joinType != CrossJoin {
 		panic("jet: join condition is nil")
 	}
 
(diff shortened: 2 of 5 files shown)


go-jet:
@OrlandoCo Thanks for your MR.

Why do we need additional join types? 
Wouldn't be simple just to add a `LATERAL` function. For instance:
```go
INNER_JOIN(
   LATERAL(
    SELECT(...).FROM(...)
   ).AS("lateral_alias"), Bool(true))
```
LATERAL function will accept selectStatement as a parameter and return SelectTable interface. Implementation will be similar as [salectTableImpl](https://github.com/go-jet/jet/blob/fdde2ab9b432f82934c5807198dfbf76521477f4/postgres/select_table.go#L11).

OrlandoCo:
Hello,

Thank you for your review, I decided to add Join methods on reference to others query builders approach to LATERAL.

I made the changes according to your observations.

OrlandoCo:
There is also missing a way to make references of the selected columns in SelectTable to use in a WHERE clause or other projections, ej:

```sql
SELECT * FROM a 
LATERAL (SELECT id AS e_id FROM b) AS e 
WHERE e.e_id=a.id
``` 
Could you please give me a hint how to implement it?

internal/jet/lateral_table.go
```diff
@@ -0,0 +1,41 @@
+package jet
+
+// LateralTable is interface for Lateral sub-queries
+type LateralTable interface {
```
go-jet, line 4: This type already exist -`SelectTable`. No need to add a new one.

internal/jet/lateral_table.go
```diff
@@ -0,0 +1,41 @@
...
+
+	projectionList := statementWithProjections.projections().fromImpl(s)
+	return projectionList.(ProjectionList)
+}
+
+func (s lateralTableImpl) serialize(statement StatementType, out *SQLBuilder, options ...SerializeOption) {
+	out.WriteString("LATERAL")
+	s.selectStmt.serialize(statement, out)
+
+	out.WriteString("AS")
```
go-jet, line 39: we can skip this part if alias is "".

internal/jet/table.go
```diff
@@ -186,3 +186,68 @@ func (t *joinTableImpl) serialize(statement StatementType, out *SQLBuilder, opti
 		t.onCondition.serialize(statement, out)
 	}
 }
+
```
go-jet, line 189: I don't understand. Whats the point of the changes in this file? 
Lateral is not a join type. Lateral logic is already implemented in the changes above (`lateralTableImpl`).

postgres/lateral_expression.go
```diff
@@ -0,0 +1,41 @@
+package postgres
+
+import (
+	"github.com/go-jet/jet/v2/internal/jet"
+)
+
+type LateralExpression interface {
+	As(alias string) LateralTable
```
go-jet, line 8: You can move `As` from this interface into bellow `LateralTable` interface and remove this one. 
Also use capital case for SQL keywords - `AS`.

postgres/table.go
```diff
@@ -13,6 +13,9 @@ type readableTable interface {
 	// Generates a select query on the current tableName.
 	SELECT(projection Projection, projections ...Projection) SelectStatement
 
+	// Creates a Lateral query on the current tableName.
+	LATERAL(table SelectTable) ReadableTable
```
go-jet, line 17: Why?

tests/postgres/select_test.go
```diff
@@ -920,6 +921,136 @@ FROM dvds.actor
...
+     actor.last_update AS "actor.last_update",
+     film_actor.actor_id AS "film_actor.actor_id",
+     film_actor.film_id AS "film_actor.film_id",
+     film_actor.last_update AS "film_actor.last_update",
+     "rFilms"."film.film_id" AS "film.film_id",
+     "rFilms"."film.title" AS "film.title",
+     "rFilms"."film.rating" AS "film.rating"
+FROM dvds.actor
+     INNER JOIN dvds.film_actor ON (actor.actor_id = film_actor.film_id)
+     LATERAL (
```
go-jet, line 1025: There is no such thing as LATERAL join. When there is no join specified, it will default to CROSS JOIN. 
This is just a CROSS JOIN between two tables, where the second is lateral query. So this will be equivalent to:
```go
.CROSS_JOIN(LATERAL(....).AS("rFilms"))
```
Also you are missing a comma here.

OrlandoCo: LATERAL are not joins and not intended to use with JOINS only, LATERAL are for subqueries, according to documentation:

>Subqueries appearing in FROM can be preceded by the key word LATERAL...
> A LATERAL item can appear at top level in the FROM list, or within a JOIN tree. In the latter case it can also refer to any items that are on the left-hand side of a JOIN that it is on the right-hand side of.

LATERAL dont defaults to CROSS JOIN.

postgres/table.go
OrlandoCo: To be used as shortcut, so it can be used like:

```go
Users.SELECT(Users.AllColumns()).FROM(Users.LATERAL(subquery...))

Users.LATERAL(subquery...).SELECT(...)
```

It can be eliminated if required.

internal/jet/table.go
OrlandoCo: To be used as shortcut, it can be deleted, please refer to below answer.

internal/jet/lateral_table.go
OrlandoCo: Maybe throw an exception, as LATERAL is for subqueries it **must** be aliased

internal/jet/lateral_table.go
OrlandoCo: As LATERAL are a especial kind of subquery that produces a derived table, I wanted to use it's own interface so maybe in the future we can add some constraints, like the accessing variables are not in order etc...

postgres/lateral_expression.go
OrlandoCo: Yeah sorry my IDE changed that...

LATERAL expression recieves a SelectStatment, but to enforce the use of ALIAS it returns LateralExpression, so it nows requires to use the AS method to return a LateralTable

tests/postgres/select_test.go
go-jet: What I wanted to say, if there was a comma (at the end of line 1024) it would be CROSS JOIN. Written this way, I don't know what LATERAL represents.

According to documentation: 
From clause is in the form:
`FROM table_reference [, table_reference [, ...]]` -- top level from list

A table reference can be a table name (possibly schema-qualified), or a derived table such as a subquery, a JOIN construct, or complex combinations of these.

Where join table is this form:
`T1 join_type T2 [ join_condition ]`

In this example. T1 is `dvds.actor`, T2 is `dvds.film_actor` and join condition is `ON (actor.actor_id = film_actor.film_id)`.

What is LATERAL? Is it part of the `join_condition`?
And we know it is not a join type.

internal/jet/lateral_table.go
go-jet: We could add it later, if there is a need for new interface.

internal/jet/lateral_table.go
go-jet: If it has to be aliased, then it is ok.

postgres/lateral_expression.go
go-jet: Ok, it makes sense.

tests/postgres/select_test.go
OrlandoCo: LATERAL represents a subquery, and yes it's missing a comma, all LATERAL subqueries must be separated by commas

OrlandoCo:
Need to review the code more, I have no idea how to implement without adding other interfaces or polluting the originals.

PR #51, opened 2020-09-03
export OrderByClause type through alias

### LINKED ISSUE #48: Dynamically construct ORDER BY
Hi, I have a question.

Based on a selection of fields, I am dynamically joining tables by conditionally calling `LEFT_JOIN` and appending the joined table's columns to a `[]mysql.Projection` which I use in the final `SELECT()`. This works pretty well.

Now I want to dynamically construct the `ORDER_BY()` based on the columns I joined, but I fail to see how. The function takes arguments of the internal type `jet.OrderByClause`, and I have not found an exported type (or interface) I can use to have a slice of these I can append to.

Did I miss something, or is there currently no way to do this? I would really prefer to avoid using subselects in the joined tables just for the ordering.

Thanks in advance!

gkdr:
enables e.g. creating a collection of clauses beforehand and dynamically adding to it.
this resolves #48.

### DIFF
--- a/mysql/delete_statement.go
+++ b/mysql/delete_statement.go
@@ -7,7 +7,7 @@ type DeleteStatement interface {
 	Statement
 
 	WHERE(expression BoolExpression) DeleteStatement
-	ORDER_BY(orderByClauses ...jet.OrderByClause) DeleteStatement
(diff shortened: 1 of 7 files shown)


go-jet:
Looks good.

PR #55, opened 2020-09-29
Add support for custom functions.

tolfino:
This allows expressions like jet.Func("FOO", String("test")) to be
emitted as FOO($1).

### DIFF
--- a/internal/jet/func_expression.go
+++ b/internal/jet/func_expression.go
@@ -801,3 +801,8 @@ func newTimestampzFunc(name string, expressions ...Expression) *timestampzFunc {
 
 	return timestampzFunc
 }
+
+// Func can be used to call an custom or as of yet unsupported function in the database.
+func Func(name string, expressions ...Expression) Expression {
+	return newFunc(name, expressions, nil)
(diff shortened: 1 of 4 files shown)


go-jet:
Thanks for the MR.

I think a better approach would be to implement custom functions rather than custom expressions.

Method signature would be:
```go
func FUNC(name string, expressions ...Expression) Expression
```
this would allow us to write:
```go
FUNC("foo") => foo()
FUNC("foo", String("name")) => foo(?)
FUNC("foo", Customer.FirstName.CONCAT(Customer.LastName)) => foo(customer.first_name || customer.last_name)
```

go-jet:
Not sure at the moment about custom operators, it might be functions like PostfixOp(), PrefixOp, or something similar.
There might be a case where there is no workaround, except to add custom expressions support, but would rather get that case as an issue, then to implement it up front.
For this issue, just a custom function support would be enough.

tolfino:
Done, please take another look.

go-jet:
Looks good. Thanks.

PR #61, opened 2020-11-16
lossless decimal type with float64 backward compatibility

crazydadz:
PR based on your decimal branch that is backward compatible with existing generated models that uses the float64 type. All the postgres tests are passing. This may breaks custom models that defined their own decimal type but paradoxically this may be a good thing if they cares about decimal precision.


go-jet:
Thanks, looks good.

PR #62, opened 2020-11-19
fix insert of decimal by changing decimal library

crazydadz:
Unfortunately [ericlagergren/decimal](https://github.com/ericlagergren/decimal) does not work as is with `jet` to write into the database. The decimal object is implementing `driver.Valuer` with a pointer receiver and `jet` is using `reflect.Indirect()` to build the arguments of the statement so the driver does not recognize the decimal object as a `Valuer`.

I've modified the tests to use [shopspring/decimal](https://github.com/shopspring/decimal) for both reading and writing into the database.


go-jet:
Nice. Good catch.

Issue #65, opened 2021-01-02
Add raw helper with parameters


Sytten:
Currently the `Raw()` function takes a string, but sometimes functions require parameters from the user. Currently you have to sanitize yourself the input and use a `fmt.Sprintf` which is prone to SQL injections.
It would be nice if we could something like: `Raw("my_function($1)", String("test"))`.

go-jet:
Is the `my_function` a custom function or it is currently unsupported SQL function?
If it is a custom function, you can use `Func`. For instance:
```Func("my_function", String("test"), Int(11))```
If it is currently unsupported SQL function, above `Func` can also be used, or we could add support for the missing function. 
_Func support is added to `develop` branch_ https://github.com/go-jet/jet/pull/55

Sytten:
Ho, Func would be ok for my use case.
Maybe a release would be good? Lets also add an example to the doc.

PR #68, opened 2021-01-11
Add RawP expression

Sytten:
Fixes https://github.com/go-jet/jet/issues/65
It takes a similar approach as `fmt.Sprintf`, let me know if you like it.

The only thing is that you can't set parents when using parameters, any idea on how to fix that?
Maybe we could add a `Parent` helper that adds the parents to a given expression.

### DIFF
--- a/internal/jet/literal_expression.go
+++ b/internal/jet/literal_expression.go
@@ -2,6 +2,7 @@ package jet
 
 import (
 	"fmt"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


go-jet:
I feel we could reuse existing `Raw` function.
`func Raw(raw string, parent ...Expression) Expression {` 
`parent ...Expression` - is something that shouldn't be exposed to the user, it is internal construct, so it safe to assume it is not used by users.

We could rename existing Raw in `internal/jet` into `RawWithParent`, and create a new one:
`func Raw(raw string, args ...interface{}) Expression {`
`args` - can be interface{}, don't see a benefit of Expression

Then serialization would be much simpler:
```go
func (n *rawExpression) serialize(statement StatementType, out *SQLBuilder, options ...SerializeOption) {
	out.WriteString(n.Raw)
        out.Args = append(out.Args, n.Args...)
}
```

This should work for MySQL, but for postgres additional change is required.
postgres dialect uses `$1`, `$2`, `$3` for argument placeholders. 
We could either use `?` placeholder as well in postgres or we could parse raw string and replace `?` with `${len(out.Arg)}`.

Sytten:
I think I was under the impression that Raw could be used within a wider select statement. That is how I use it anyway because the Func is not merged to master.

So for me it was really important that the number in the arguments match because they might be just a subset of the arguments of the statement. That is why I took as input Expression so I could serialize them and they would get the correct number. I don't really mind how we parse the string, either $1 or ?.

So please tell me if you think Raw should be used within other statements or not.

go-jet:
Could you share example of input Expression you had in mind?

Sytten:
Well for example I could have used raw to emulate the values expression on my other issue, but it would not have worked if we don't rewrite the query for the arguments to match.
Its just a matter of if the raw can or cannot be used to write part of a query.

go-jet:
Agreed, number of arguments should match.
Because Raw will return Expression it will be able to use it as part of the bigger query.
For instance:
```go
Actor.Name.EQ(StringExp(
    Raw("my_function($1, $2, $3)", 11, "John", 22.33),
)
```

I feel that Raw should look like native golang/sql queries. Meaning it will work with basic golang types, and should not be used with jet expressions.
If some of the arguments is column or expression it should be written directly in query.
For instance:
```
Raw("actor.name || $1", "Doe")   // rather then Raw("$1 || $2", Actor.Name, String("Doe"))
```

Sytten:
It would be nice, but I feel a first iteration with `$1` or `?` is a good first step to merge and release (it's been a while since a release).

go-jet:
This issue is resolved with [2.5.0 release](https://github.com/go-jet/jet/wiki/Expressions#raw-operator), so I'm gonna close this PR.

PR #86, opened 2021-07-29
Export GroupByClause type through alias

mlaflamm:
Allows to dynamically create group by clauses.


go-jet:
Thanks for the PR. Looks good.

Issue #91, opened 2021-09-14
`QueryContext` panic when the scanned value is nil and the destination is a slice of primitive


mlaflamm:
I have a custom model that look like this:

```
var dest struct {
  Count int `sql:"primary_key"`
  ID    []string
}
```

When the id column is null like this, `QueryContext` panic with an error `call of reflect.Value.Type on zero Value`

```
count|id|
-----+--+
    0|  |
```

I’ve tracked the issue in the function `scanContext.rowElemValuePtr` of file `qrm/scan_context.go`. “ID” is a slice of primitive and since the value is nil, there is not enough type information to create the slice by reflection.

I am working on a PR.

PR #92, opened 2021-09-14
Fix panic when scanning a nil value and the destination is a slice of primitive type

mlaflamm:
Fix for #91

### DIFF
--- a/qrm/scan_context.go
+++ b/qrm/scan_context.go
@@ -222,19 +222,23 @@ func (s *scanContext) rowElem(index int) interface{} {
 	return value
 }
 
-func (s *scanContext) rowElemValuePtr(index int) reflect.Value {
+func (s *scanContext) rowElemValuePtr(index int) *reflect.Value {
 	rowElem := s.rowElem(index)
 	rowElemValue := reflect.ValueOf(rowElem)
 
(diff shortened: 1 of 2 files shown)


qrm/scan_context.go
```diff
@@ -222,19 +222,23 @@ func (s *scanContext) rowElem(index int) interface{} {
...
 
-	newElem := reflect.New(rowElemValue.Type())
-	newElem.Elem().Set(rowElemValue)
-	return newElem
+	if rowElemValue.IsValid() {
+		newElem := reflect.New(rowElemValue.Type())
+		newElem.Elem().Set(rowElemValue)
+		return &newElem
+	}
+	return nil
```
mlaflamm, line 243: I could returns `reflect.Value{}` instead of `nil` and not change the signature of this function. We need to verify for `IsValid` instead of `!= nil` in the calling function.

go-jet: Agreed, I think that would be a better approach.

go-jet: Agree, I think that would be a bit cleaner approach.

mlaflamm: Updated

go-jet: LGTM.

Issue #95, opened 2021-09-21
Special characters are not working in postgres password(for generator)


masterkidan:
It looks like there's a bug in the underlyign pq lib that is being used for making a connection to the postgres server. 
https://github.com/lib/pq/issues/787#issuecomment-419621490

The recommendation is to use the postgres connection string and url encoded password.

PR #96, opened 2021-09-21
fix: Fixing up postgres connection strings

masterkidan:
This is to fix issue #95


go-jet:
Could you open PR for `develop` branch.  There are some changes made in generator that does not exist on master.  
`-dsn` flag was added with  https://github.com/go-jet/jet/pull/89.
I guess we could still replace flags with connection string for this function.

masterkidan:
Updated to develop (hopefully I updated the above correctly).

go-jet:
LGTM, but it seems there are some rebase conflicts.

masterkidan:
Can you clarify where the conflicts are? I don't see them in the diff.

go-jet:
It is probably because of merge commit a5c3d545fe376b87fc4eae36ca208fbf9a0fafad. But it is ok, squash and merge will work as well.

Issue #111, opened 2021-12-14
Generator no longer supports MySQL 5.7


kblomster:
**Describe the bug**
I'm not sure if this strictly speaking a bug since AFAIK there's no list of officially supported DBMS versions, but it did work in 2.5.0 and isn't working in 2.6.0 so here we are.

Attempting to generate models for any schema on MySQL 5.7 results in the following database error:

```
Retrieving database information...
jet: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'primaryKeys AS (
        SELECT k.column_name
        FROM information_schema.table_constraint' at line 1
```

**Environment (please complete the following information):**
 - OS: macosx 12.0.1
 - Database: MySQL 5.7
-  Database driver: mysql
 - Jet version: 2.6.0

**Expected behavior**
I expect model generation to work, as it did in 2.5.0.

**Additional information**
The error appears to originate in `GetTableColumnsMetaData()` in `jet/generator/mysql/query_set.go`; it attempts to run a query that uses a CTE (`WITH primaryKeys AS (...`) and MySQL 5.7 doesn't support CTE's - which is one of the many reasons I despise it, but I have to use it for now, because of reasons.

It doesn't seem too difficult to rewrite the query to avoid the CTE (at the cost of some readability), so if you're willing to accept that tradeoff to keep supporting MySQL 5.7 I'd be very happy. I could make a PR?

go-jet:
Yeah, it is a bug. Generator should work with MySQL 5.7. 
Please use `develop` branch for PR.

kblomster:
Tested the develop branch now and it works fine again. Thanks for the quick response!

PR #112, opened 2021-12-14
generator: support MySQL 5.7

kblomster:
MySQL 5.7 does not support CTE's. Refactor the table column metadata query to avoid the CTE. Fixes regression introduced in 8864667f4772; see also issue #111.

Some of the MySQL tests are failing on 5.7 but not any of the generator ones, just some SELECT ones that use syntax that MySQL 5.7 doesn't have, so that's only to be expected. As a side note, to import `test_sample.sql` from the testdata repo into 5.7 with (AFAIK) default settings, I had to add the line `SET SQL_MODE='ALLOW_INVALID_DATES';` or it'd complain about an invalid default for the column `timestamp_ptr`. I'm not sure if I can/should do anything about that, so I left it as is.


go-jet:
Perfect. Looks good.

PR #138, opened 2022-04-13
fix -ignore-tables, -ignore-enums and -ignore-views when -dsn is present

nkonin:


go-jet:
Thanks, @nkonin . LGTM.

PR #159, opened 2022-06-03
mysql: export some conditional functions

### LINKED ISSUE #158: COALESCE missing in MySQL
**Describe the bug**
The postgres dialect exports `COALESCE` in `postgres/functions.go`, but the MySQL dialect lacks it. It's a pretty ubiquitous function, so I believe this is probably a bug.

**Code snippet**
```golang
q := Film.SELECT(
    Film.Title,
    IntExp(COALESCE(Film.Length, Int(0))).AS("length_not_null")
)
```

**Expected behavior**
I expect `COALESCE` to be available the MySQL dialect.

As a minor feature request, I think it'd be pretty nice to have the functions listed somewhere on the wiki. Some of them are already on the "Expressions" page, and to be fair a lot of them are just the exact same keywords as in SQL, but when you're new to the library it really helps to just know for sure that they do exist (and especially so when it comes to Jet-specific variants like `SUMf()` - it shows up in examples, but again, a reference would be nice).

Having typed this out I now realize that in the time it took I could just have submitted a PR instead. :/

kblomster:
`COALESCE`, `NULLIF`, `GREATEST` and `LEAST` already existed and were available in the postgres dialect, but not in MySQL. `GREATEST` and `LEAST` behave differently with nulls than they do in Postgres, but I don't think that matters to Jet.

Fixes #158.


go-jet:
Thanks, looks good.

Issue #160, opened 2022-06-04
Add postgres JSON cast type


wexder:
**Is your feature request related to a problem? Please describe.**
We are using a lot of `json` fields in our postgres database. And it's hard to do the UPSERT with `json` column type.
Currect solution I found is double casting the json string. 
```
  insertStm := table.INSERT(table.AllColumns).MODEL(table).
  ON_CONFLICT(table.SomeColumn).DO_UPDATE(
  postgres.SET(
    table.JsonColumn.SET(postgres.StringExp(postgres.CAST(postgres.String(*data)).AS("json"))),
    ),
```

**Describe the solution you'd like**
Add `postgres.Json()` func that would return `StringExpression`.

wexder:
I've create PR for this https://github.com/go-jet/jet/pull/161

go-jet:
Hi @wexder. Yeah, it makes sense to have a json literal.

go-jet:
Support added with v2.9.0 release.

PR #161, opened 2022-06-04
Add postgres json literal

wexder:
Json types are crucial for postgres. This PR add support to easily cast golang string as postgres json type.

Solves https://github.com/go-jet/jet/issues/160.


go-jet:
Thanks, looks good.

PR #171, opened 2022-08-18
add missing SET method to ColumnTimez interface

nkonin:
it has been already implemented, just missing from interface


go-jet:
Thanks. :+1:

PR #175, opened 2022-09-06
fix: Add omitted reserved word 'RIGHT' on postgresql

Hoonyyhoon:
https://www.postgresql.org/docs/current/sql-keywords-appendix.html

Seems like reserved word "RIGHT" omitted in the `reservedWords` in the postgresql dialect


go-jet:
Thanks.

PR #179, opened 2022-09-29
generate slice types on the model

samnguyen-outreach:


go-jet:
Hi @samnguyen-outreach. Its a bit more complicated with arrays than for the other sql types. Array types are allowed to contain NULL values. So for those arrays destination will be `[]*T`, rather than `[]T`. That information can't be retrieved from the information schema, so we can't make assumptions during file generation. 
Also, setting a type to be `[]T`, would cause query result mapping to fail, because `qrm` doesn't know how to parse arrays. Array type would need to implement `Scanner` and `Valuer` interface so that `qrm` can parse those array values. There are already different array implementations (like `pq`, `pgx`). So which one to choose?
Because we can't make the assumption that will work for everyone, developers need to specify their array type explicitly. You can achieve that using [generator customization](https://github.com/go-jet/jet/wiki/Generator#generator-customization).

samnguyen-outreach:
Thanks for the response, I will go toward customizing the generator then.

PR #184, opened 2022-10-24
Potential security flaw

yngfoxx:
I am using go-jet as a go library for an application that uses dynamic database credentials and noticed this potential security risk that exposes the whole database credentials (though it is timed, it is still a risk)

### DIFF
--- a/generator/postgres/postgres_generator.go
+++ b/generator/postgres/postgres_generator.go
@@ -69,7 +69,7 @@ func GenerateDSN(dsn, schema, destDir string, templates ...template.Template) (e
 }
 
 func openConnection(dsn string) *sql.DB {
-	fmt.Println("Connecting to postgres database: " + dsn)
+	fmt.Printf("Connecting to postgres database: %v...", dsn[:15])
 
 	db, err := sql.Open("postgres", dsn)
 	throw.OnError(err)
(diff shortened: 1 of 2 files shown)


generator/postgres/postgres_generator.go
```diff
@@ -69,7 +69,7 @@ func GenerateDSN(dsn, schema, destDir string, templates ...template.Template) (e
 }
 
 func openConnection(dsn string) *sql.DB {
-	fmt.Println("Connecting to postgres database: " + dsn)
+	fmt.Printf("Connecting to postgres database: %v...", dsn[:15])
```
go-jet, line 72: I think we can just remove `+ dsn` part.

go-jet:
Although jet generator is intended to be used on local db instance only, we can't rule out possibility that someone will use it on a production db, so yeah it is probably the best to remove `dsn` from log completely.

generator/postgres/postgres_generator.go
go-jet: The same for `mysql`.

yngfoxx: Yeah that would be the best

go-jet:
Thanks, looks good.

Issue #79, opened 2021-05-24
Support for dynamic schema names


mickeyreiss:
At the moment, the schema name is baked into the generated code. It would be helpful if the schema name could be globally overridden, so that a compiled program could be configured to hit different databases in different environments without changing code (or adding `FromSchema` to all queries).

go-jet:
What would be the use case? To set a default schema name during server startup and then during the server lifespan only that schema will be used?
If thats the case, there is another workaround. It is possible to override generated sql builder table objects with a different table objects(with different schema name).
For instance:
```go
import "github.com/go-jet/jet/v2/examples/quick-start/.gen/jetdb/dvds/table"

func serverStartup(schemaName string){
....
table.Artist = table.Artist.FromSchema(schemaName)
table.Album = table.Album.FromSchema(schemaName)
table.Customer = table.Customer.FromSchema(schemaName)
...
}
```
Because these table objects are global variables there is no need to change a code or add FromSchema to all queries.

mickeyreiss:
The use case is different schema names in dev, staging, production.

This workaround seems like it will work, albeit still requiring one extra line of code when a new table is added.

gkdr:
funny enough, we've had our own fork with such a functionality hacked into the `NewTable()` function: it reads an env variable and checks if there is an alias for the current schemaname, see https://github.com/unwired/jet/blob/5426d04726da9c209fce94ce8f925d9591415e4c/mysql/table.go#L88

that was from before the schema alias functionality from the newest release (afaik), so i believe there was no other way at that point in time :slightly_smiling_face: would you consider such a functionality in any form, or is the workaround you provided the "official" way now?

go-jet:
Yes, `FromSchema` method was added to the latest release. In the feature, new auto generated global function might be added, to automate schema rename of all global table objects.
Something like: `table.UseSchema(schemaName)` to replace above `serverStartup` function.

realbucksavage:
@mickeyreiss @go-jet I can think of a possibility to let the generated MySQL table code have a blank `schemaName`. Since there is already a check for empty schema names at https://github.com/realbucksavage/jet/blob/master/internal/jet/table.go#L78, the query will go to the schema specified in DSN used to create the database connection.

For example, doing `jet -source mysql -dsn "..." -path ./gen -skip-schema` will get us

```go
var MyTable = newMyTableTable("", "my_table", "")
```

I can submit a PR if this sounds like worth a shot.

realbucksavage:
[...]

PR #194, opened 2022-12-01
Added a global `UseSchema` method to table generation

realbucksavage:
Adds a convenience method that can be used to change the schema name for all generated table instances. This could be a potential solution to #79 where you would want to use schema names dynamically in your program.

This could let you do things like this:

```go
package main

import (
  // ...
  "mymodule.com/gen/dvds/table"
)

func main() {
  // ...
  schemaName := determineSchemaName()
  table.SetDvdsSchema(schemaName)
}
```

Passing an empty string to `SetSchema` would cause the queries to run in the schema specified in the DSN used to create DB connection.


generator/template/process.go
```diff
@@ -61,6 +63,52 @@ func processSQLBuilder(dirPath string, dialect jet.Dialect, schemaMetaData metad
 	processTableSQLBuilder("table", sqlBuilderPath, dialect, schemaMetaData, schemaMetaData.TablesMetaData, sqlBuilderTemplate)
 	processTableSQLBuilder("view", sqlBuilderPath, dialect, schemaMetaData, schemaMetaData.ViewsMetaData, sqlBuilderTemplate)
 	processEnumSQLBuilder(sqlBuilderPath, dialect, schemaMetaData.EnumsMetaData, sqlBuilderTemplate)
+	processTableSQLBuilderSetSchema(sqlBuilderPath, schemaMetaData, sqlBuilderTemplate)
```
go-jet, line 66: This method will be also useful for views as well. Not just tables.

go-jet: I don't think we need another method, we can reuse `processTableSQLBuilder`, it used for both tables and views.

generator/template/process.go
```diff
@@ -61,6 +63,52 @@ func processSQLBuilder(dirPath string, dialect jet.Dialect, schemaMetaData metad
...
+		return
+	}
+
+	err := utils.EnsureDirPath(dirPath)
+	throw.OnError(err)
+
+	fmt.Println("Generating global `SetSchema` method")
+	schemaIdentifier := utils.ToGoIdentifier(schemaMetadata.Name)
+
+	funcPath := path.Join(dirPath, builders[0].Path)
```
go-jet, line 92: file can be  named just `table.go` and `view.go`. You can reuse `fileTypes` inside `processTableSQLBuilder`.

generator/template/process.go
realbucksavage: Great idea. I'm thinking of a `generateSetSchema` method that can be called from `processTableSQLBuilder`. This can be used to generate `table.go` and `view.go` by using `fileTypes` variable that is already there. Thoughts?

generator/template/process.go
realbucksavage: Updated in 7db99b1

generator/template/process.go
```diff
@@ -169,9 +172,40 @@ func processTableSQLBuilder(fileTypes, dirPath string,
...
+	if len(generatedBuilders) > 0 {
+		generateSetSchema(dirPath, fileTypes, schemaMetaData, generatedBuilders)
 	}
 }
 
+func generateSetSchema(dirPath, fileTypes string, schemaMetadata metadata.Schema, builders []TableSQLBuilder) {
+
+	basePath := path.Join(dirPath, builders[0].Path)
+	err := utils.EnsureDirPath(basePath)
+	throw.OnError(err)
```
go-jet, line 188: Caller function already insured directory exist.

tests/mysql/generator_test.go
```diff
@@ -309,6 +312,36 @@ func newActorTableImpl(schemaName, tableName, alias string) actorTable {
...
+var tableSetSchemaFile = `
+//
+// Code generated by go-jet DO NOT EDIT.
+//
+// WARNING: Changes to this file may cause incorrect behavior
+// and will be lost if the code is regenerated
+//
+
+package table
+
```
go-jet, line 324: Go doc comment to describe what this function does.

tests/mysql/generator_test.go
```diff
@@ -309,6 +312,36 @@ func newActorTableImpl(schemaName, tableName, alias string) actorTable {
...
+func SetDvdsSchema(schema string) {
```
go-jet, line 325: Not sure about name. If schema name appears in the function name, it feels like we are setting something in the schema. I was thinking of `SetSchema` or `UseSchema`.

realbucksavage: `UseSchema` makes much more sense. Thanks for the input.

tests/mysql/generator_test.go
realbucksavage: Resolved in 04c1a51

generator/template/process.go
realbucksavage: Resolved in 04c1a51

generator/template/process.go
```diff
@@ -169,9 +172,38 @@ func processTableSQLBuilder(fileTypes, dirPath string,
...
+	if len(generatedBuilders) > 0 {
+		generateUseSchemaFunc(dirPath, fileTypes, schemaMetaData, generatedBuilders)
 	}
 }
 
+func generateUseSchemaFunc(dirPath, fileTypes string, schemaMetadata metadata.Schema, builders []TableSQLBuilder) {
+
+	basePath := path.Join(dirPath, builders[0].Path)
+
+	schemaIdentifier := utils.ToGoIdentifier(schemaMetadata.Name)
```
go-jet, line 188: My [comment](https://github.com/go-jet/jet/pull/194#discussion_r1038840362) was mainly about using the current schema identifier in the function name. Because we are replacing 'dvds' as schema name not setting. So I'm thinking it's better to go without a schema identifier in the function name completely.

generator/template/process.go
```diff
@@ -169,9 +172,38 @@ func processTableSQLBuilder(fileTypes, dirPath string,
...
+	}
+
+	if len(generatedBuilders) > 0 {
+		generateUseSchemaFunc(dirPath, fileTypes, schemaMetaData, generatedBuilders)
 	}
 }
 
+func generateUseSchemaFunc(dirPath, fileTypes string, schemaMetadata metadata.Schema, builders []TableSQLBuilder) {
+
+	basePath := path.Join(dirPath, builders[0].Path)
```
go-jet, line 186: tableSQLBuilderPath can be passed from the caller.

generator/template/process.go
realbucksavage: @go-jet `tests/mysql/update_test.go` has a dot import for both `.gentestdata/mysql/dvds/table` and `.gentestdata/mysql/test_sample/table`. Simply generating `UseSchema` would cause a "Function `UseSchema` redeclared in this package" error. This was my initial reasoning behind using the schema identifier in the generated method.

generator/template/process.go
go-jet: Ignore this. I just noticed tableSQLBuilderPath is inside a loop.

generator/template/process.go
go-jet: Aha, I see. We can extract test related to `.gentestdata/mysql/dvds/table` into separate `update_dvds_test.go` file. There is just one test there - `TestUpdateWithJoin`.

realbucksavage: Ah this makes sense. Perfect!

generator/template/process.go
realbucksavage: @go-jet good point actually. `generateUseSchemaFunc` can simply accept a single `TableSQLBuilder`, passed down by `processTableSQLBuilder`. The invocation could look like `generateUseSchemaFunc(dirPath, fileTypes, generatedBuilders[0])`.

generator/template/process.go
realbucksavage: @go-jet I'm wondering why CircleCI tests are failing. These tests work just fine when run with `go run ./...`.

go-jet: It seems that tableMetaData slice is unordered, so on different machines slightly different file is constructed. Adding `ORDER BY table_name` [here](https://github.com/go-jet/jet/blob/c5e3d1ee7505a16b0fcb4fcf09d89e64ca5d0adf/generator/postgres/query_set.go#L18) and [here](https://github.com/go-jet/jet/blob/c5e3d1ee7505a16b0fcb4fcf09d89e64ca5d0adf/generator/mysql/query_set.go#L19), should probably fix the tests.

go-jet: Great. It looks good. :+1:

go-jet (approved):

PR #197, opened 2022-12-16
Allow custom types as indexes when joining tables

emarj:
See #198 

The change is quite simple. We check if the field is tagged as primary key, if so, we add it as an index regardless of its type.

### DIFF
--- a/qrm/scan_context.go
+++ b/qrm/scan_context.go
@@ -187,20 +187,7 @@ func (s *ScanContext) getGroupKeyInfo(
 		field := structType.Field(i)
 		fieldType := indirectType(field.Type)
 
-		if !isPrimaryKey(field, primaryKeyOverwrites) {
-			if !isSimpleModelType(fieldType) {
-				if fieldType.Kind() != reflect.Struct {
-					continue
-				}
-
-				subType := s.getGroupKeyInfo(fieldType, &field, typeVisited)
-
-				if len(subType.indexes) != 0 || len(subType.subTypes) != 0 {
-					ret.subTypes = append(ret.subTypes, subType)
-				}
-			}
-		} else {
-
+		if isPrimaryKey(field, primaryKeyOverwrites) {
 			newTypeName, fieldName := getTypeAndFieldName(typeName, field)
 
 			index := s.typeToColumnIndex(newTypeName, fieldName)
@@ -211,6 +198,13 @@ func (s *ScanContext) getGroupKeyInfo(
 
 			ret.indexes = append(ret.indexes, index)
 
+		} else if fieldType.Kind() == reflect.Struct {
+
+			subType := s.getGroupKeyInfo(fieldType, &field, typeVisited)
+
+			if len(subType.indexes) != 0 || len(subType.subTypes) != 0 {
+				ret.subTypes = append(ret.subTypes, subType)
+			}
 		}
 	}
 
--- a/qrm/scan_context.go
+++ b/qrm/scan_context.go
@@ -187,28 +187,30 @@ func (s *ScanContext) getGroupKeyInfo(
 		field := structType.Field(i)
 		fieldType := indirectType(field.Type)
 
-		if !isSimpleModelType(fieldType) {
-			if fieldType.Kind() != reflect.Struct {
-				continue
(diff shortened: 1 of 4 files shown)


emarj:
The change is quite simple. We check if the field is tagged as primary key, if so, we add it as an index, regardless of its type.

go-jet:
Good catch. :100:  Could you also include one test in `./tests/postgres` (or mysql).

qrm/scan_context.go
```diff
@@ -187,28 +187,30 @@ func (s *ScanContext) getGroupKeyInfo(
 		field := structType.Field(i)
 		fieldType := indirectType(field.Type)
 
-		if !isSimpleModelType(fieldType) {
-			if fieldType.Kind() != reflect.Struct {
-				continue
-			}
+		if !isPrimaryKey(field, primaryKeyOverwrites) {
+			if !isSimpleModelType(fieldType) {
+				if fieldType.Kind() != reflect.Struct {
```
go-jet, line 192: Maybe we can reorder this condition, to avoid two ifs with negations. 
```golang
if isPrimaryKey(...) {
...
} else !isSimpleModelType(fieldType) {
...
}
```

go-jet: Actually it can be just:

```golang
if isPrimaryKey(...) {
...
} else fieldType.Kind() == reflect.Struct {
...
}
```

emarj: Yeas you are right. Done!

emarj:
I added the test. Unfortunately I was not able to properly configure the test environment on my local machine. I hope it passes on CI.

tests/postgres/select_test.go
```diff
@@ -658,6 +660,98 @@ ORDER BY city.city_id, address.address_id, customer.customer_id;
...
+WHERE (city.city = 'London'::text) OR (city.city = 'York'::text)
+ORDER BY city.city_id, address.address_id, customer.customer_id;
+`, "London", "York")
+
+	err := stmt.Query(db, &dest)
+
+	require.NoError(t, err)
+	require.Equal(t, len(dest), 2)
+	testutils.AssertJSON(t, dest, `
+[
```
go-jet, line 714: It seems that tests fails because, json doesn't match. Expected json:

```json
[
	{
		"CityID": {
			"Int64": 312,
			"Valid": true
		},
		"CityName": "London",
		"Customers": [
			{
				"CustomerID": {
					"Int64": 252,
					"Valid": true
				},
				"LastName": "Hoffman",
				"Address": {
					"AddressID": {
						"Int64": 256,
						"Valid": true
					},
					"AddressLine": "1497 Yuzhou Drive"
				}
			},
			{
				"CustomerID": {
					"Int64": 512,
					"Valid": true
				},
				"LastName": "Vines",
				"Address": {
					"AddressID": {
						"Int64": 517,
						"Valid": true
					},
					"AddressLine": "548 Uruapan Street"
				}
			}
		]
	},
	{
		"CityID": {
			"Int64": 589,
			"Valid": true
		},
		"CityName": "York",
		"Customers": [
			{
				"CustomerID": {
					"Int64": 497,
					"Valid": true
				},
				"LastName": "Sledge",
				"Address": {
					"AddressID": {
						"Int64": 502,
						"Valid": true
					},
					"AddressLine": "1515 Korla Way"
				}
			}
		]
	}
]
```

emarj: Yes, it was silly on my part. I am so used to `null.Int` by `github.com/guregu/null` that I forgot about the JSON marshaling of `sql.NullInt64`. I also added another test by using the aforementioned `null.Int`.

I was able to configure the environment to run some test, even tough not everything works. I had to pass trough some hoops to make it work on my M1 Mac. In my case `platform: linux/amd64` was needed in `docker-compose.yaml`.

go-jet: Yeah, I've heard M1 Mac tends to have problems with docker. Maybe we can include your `docker-compose.yaml` changes. M1 is getting popular, someone else might stumble on the same issue. It can be a separate PR.

go-jet:
Perfect. LGTM.

Issue #209, opened 2023-01-25
How to customise mutableColumns to read-only generated columns in PostgreSQL?


rangzen:
Hello,
We use generated columns with PostgreSQL, see [Generated Columns](https://www.postgresql.org/docs/15/ddl-generated-columns.html).
We would like to exclude these columns from the generated mutableColumns list.
Can we use something like [Generator customization](https://github.com/go-jet/jet/wiki/Generator#generator-customization)?
Is this possible without adding fields to `metadata.Table` or `metadata.Column`? Without changing `DefaultTableModel(tableMetaData metadata.Table) TableModel` to an interface to reimplement `(t table) MutableColumns() []Column`?
Thank you!

go-jet:
Hi @rangzen,

Currently it is not possible to exclude columns from `AllColumns` or `MutableColumns` lists.
Workaround is to create new `ColumnList`s and use those instead. For instance:

```golang
var PeopleAllColumns     = People.AllColumns.Except(People.HeightIn)
var PeopleMutableColumns = People.MutableColumns.Except(People.HeightIn)
```

In terms of implementation, I suppose adding a new `Skip' flag would be enough to signal that the column should not appear in table sql builder struct.

mlaflamm:
I just want to clarify that Postgres generated columns are read only. The goal would be to keep them in `AllColumns` and the table struct but to exclude them from `MutableColumns`.

go-jet:
Aha, I see. In that case implementation can be simplified. We can retrieve additional column `is_generated` from `information_schema.columns` and then in `MutableColumns() []Column` filter out primary key and generated columns.

The only concern is this might affect existing users. But I think it should be fine, because generated columns can't be used in either insert or update statements, so existing users are probably already excluding generated columns.

rangzen:
@go-jet PR and data pushed. Not sure how you deal with the submodule.

rangzen:
Yep, CircleCI need first the link to the data submodule
`panic: open ./testdata/init/cockroach/test_generated.sql: no such file or directory`

go-jet:
Yeah, when `jet-test-data` PR is merged, `jet` PR build will be fixed. For local development you can use makefile command - `make checkout-testdata`, in tests directory, to pull latest submodule master commit.

PR #210, opened 2023-01-30
fix generated columns in mutable columns for PostgreSQL

rangzen:
Fix https://github.com/go-jet/jet/issues/209


rangzen:
Need data from https://github.com/go-jet/jet-test-data/pull/3

tests/postgres/generated_test.go
```diff
@@ -0,0 +1,15 @@
+package postgres
+
+import (
+	. "github.com/go-jet/jet/v2/tests/.gentestdata/jetdb/test_generated/table"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
```
go-jet, line 8: This tests are intended to be used with a live db. For generated column tests we can write a tests to test simple SELECT, INSERT or UPDATE query. You can then test mutable columns by comparing `.Sql()` or `.DebugSql()` query using `testutils.AssertStatementSql` or `testutils.AssertDebugStatementSql`.  For example - [check](https://github.com/go-jet/jet/blob/ce3c622fa4f8b6f7a18a8c84a386f710e6bef430/tests/postgres/sample_test.go#L17).
Result of the db query is not of importance, just error check would be enough.

The test can be added in existing `sample_test.go`.

rangzen: SQL and test migrated and updated.

go-jet:
LGTM. I've merged test data PR. To fix this PR build you'll need to push one more commit, with updated sub-module master head. You can just call `make checkout-testdata` and then push changes.

rangzen:
CircleCI didn't take the new link to jet-test-data. Did I misse something?

rangzen:
I cannot restart the CI myself.

go-jet:
Ahh, sorry I missed this. `postgres` tests are run on two databases `postgres` and `cockrouchdb`. pg tests are successfully, but cockrochdb are failing. To fix the tests we will have to add a the same table to [cockroach test init data](https://github.com/go-jet/jet-test-data/blob/master/init/cockroach/test_sample.sql).

rangzen:
https://github.com/go-jet/jet-test-data/pull/4

generator/postgres/query_set.go
```diff
@@ -44,6 +44,7 @@ WITH primaryKeys AS (
 )
 SELECT column_name as "column.Name", 
 	   is_nullable = 'YES' as "column.isNullable",
+	   is_generated = 'ALWAYS' as "column.isGenerated",
```
go-jet, line 47: We will need to update this condition since `cockroachdb` doesn't use `ALWAYS` but `YES` as a constant.

rangzen: Do you have an example in the code where we send one way for PostgreSQL and another way for CockroachDB?

go-jet: This query is used for both databases. Cockroach shares interfaces with postgres, but there are occasionally some differences. So we just have to add: `is_generated = 'ALWAYS' or is_generated = 'YES'`

rangzen: Seems ok in local tests.

go-jet: Yep, looks good.

rangzen:
@go-jet Thank you 🥳

PR #235, opened 2023-05-01
redshift combatibility added to generator

stefano-palermo:


stefano-palermo:
If you don't like the "SELECT version()" query it can be changed to accept a "redshift" source and pass the isRedshift flag through to GenerateDSN().  
This way it's supported through "postgres" source.  Whichever is preferable.

go-jet:
Hi @stefano-palermo,
Is the difference between postgres and redshift only in enums support?

stefano-palermo:
No there's columns missing from the information_schema tables as well as a fair amount of functions that are not available in Redshift as well.

go-jet:
Aha, I see.
What do you think would be the best way to include Redshift into tests, since Redshift is closed source? Using PostgreSQL 8.0.2? Or maybe create Amazon account? Is there is a free tier on Amazon for Redshift?

generator/postgres/postgres_generator.go
```diff
@@ -60,7 +60,18 @@ func GenerateDSN(dsn, schema, destDir string, templates ...template.Template) (e
 		generatorTemplate = templates[0]
 	}
 
-	schemaMetadata := metadata.GetSchema(db, &postgresQuerySet{}, schema)
+	var isRedshift bool
+	err = db.QueryRow("SELECT version() LIKE '%Redshift%';").Scan(&isRedshift)
+	if err != nil {
```
go-jet, line 65: No need for if.

generator/postgres/query_set.go
```diff
@@ -89,3 +92,69 @@ ORDER BY n.nspname, t.typname, e.enumsortorder;`
 
 	return result
 }
+
```
go-jet, line 95: This function is the same as for postgres, I think we can just `return postgresQuerySet.GetTablesMetaData(db, schemaName, tableType)`

generator/postgres/query_set.go
```diff
@@ -89,3 +92,69 @@ ORDER BY n.nspname, t.typname, e.enumsortorder;`
...
+	_, err := qrm.Query(context.Background(), db, query, []interface{}{schemaName, tableName}, &columns)
+	throw.OnError(err)
+
+	return columns
+}
+
+// GetEnumsMetaData
+// redshift does not support enums
+func (r redshiftQuerySet) GetEnumsMetaData(db *sql.DB, schemaName string) []metadata.Enum {
+	return []metadata.Enum{}
```
go-jet, line 159: return nil

generator/postgres/query_set.go
stefano-palermo: Yes but it might be a good idea to keep it independent in case future changes to the postgres query breaks it.

generator/postgres/postgres_generator.go
stefano-palermo: What do you suggest to do to decide which query set to use then?

generator/postgres/query_set.go
stefano-palermo: Updated.

stefano-palermo:
> Aha, I see. What do you think would be the best way to include Redshift into tests, since Redshift is closed source? Using PostgreSQL 8.0.2? Or maybe create Amazon account? Is there is a free tier on Amazon for Redshift?

This is tricky.  There is no free tier, only a 2 month free trial currently.  Possibly using Postgres 8 is the best way.

generator/postgres/postgres_generator.go
stefano-palermo: Updated.

go-jet:
> This is tricky. There is no free tier, only a 2 month free trial currently. Possibly using Postgres 8 is the best way.

If we are going with Postgers 8 for tests, the next step would be to add  new [docker image](https://github.com/go-jet/jet/blob/master/tests/docker-compose.yaml). Hopefull we can reuse some of the existing test databases for [postgres](https://github.com/go-jet/jet-test-data/tree/master/init/postgres). It seems that `northwind` schema doesn't have enums. And then to test redshift we can run only postgres tests that are using `northwind` schema.

generator/postgres/query_set.go
go-jet: In case of future breaking changes there will be tests that will catch if something breaks.

stefano-palermo: Updated.

stefano-palermo:
> > This is tricky. There is no free tier, only a 2 month free trial currently. Possibly using Postgres 8 is the best way.
> 
> If we are going with Postgers 8 for tests, the next step would be to add new [docker image](https://github.com/go-jet/jet/blob/master/tests/docker-compose.yaml). Hopefull we can reuse some of the existing test databases for [postgres](https://github.com/go-jet/jet-test-data/tree/master/init/postgres). It seems that `northwind` schema doesn't have enums. And then to test redshift we can run only postgres tests that are using `northwind` schema.

Added an image that attempts to be compatible with Redshift.

PR #267, opened 2023-08-20
Fix typo in README

ryym:
An example in README uses two destination paths: `gen/` and `.gen/` . This is confusing and the latter seems correct according to the actual code, so I fixed it.

https://github.com/go-jet/jet/blob/33333585e9571ee4375c1a6ee99acff129c801ba/examples/quick-start/quick-start.go#L16


go-jet:
Thanks. :+1:

PR #274, opened 2023-09-18
10x mysql db info querying speedup

josephbuchma:
Optimisation consists of two parts:
- faster SQL query
- leverage concurrency

Sample run:
```sh
$ jet -source=mysql -dsn="$MYSQL_USER:$MYSQL_PASS@tcp(localhost:$MYSQL_PORT)/facility" -path=./gen
Connecting to MySQL database...
Retrieving database information...
        FOUND 148 table(s), 162 view(s), 13 enum(s)
Destination directory: gen/facility
Cleaning up destination directory...
Generating table model files...
Generating view model files...
Generating enum model files...
Generating table sql builder files
Generating view sql builder files
Generating enum sql builder files
```

Results:
- master: 4m9s
- with improved SQL:  42s
- with improved SQL + concurrency: 21s

Tested with MySQL 5.7 running locally in Docker.

### DIFF
--- a/generator/mysql/query_set.go
+++ b/generator/mysql/query_set.go
@@ -39,27 +39,36 @@ ORDER BY table_name;
 
 func (m mySqlQuerySet) GetTableColumnsMetaData(db *sql.DB, schemaName string, tableName string) ([]metadata.Column, error) {
 	query := `
-SELECT COLUMN_NAME AS "column.Name", 
-	IS_NULLABLE = "YES" AS "column.IsNullable",
-	columns.COLUMN_COMMENT as "column.Comment",
-	(EXISTS(
-		SELECT 1
+SELECT
+		col.COLUMN_NAME AS "column.Name",
+		col.IS_NULLABLE = "YES" AS "column.IsNullable",
+		col.COLUMN_COMMENT AS "column.Comment",
+		pk.IsPrimaryKey AS "column.IsPrimaryKey",
+		IF (col.COLUMN_TYPE = 'tinyint(1)',
+				'boolean',
+				IF (col.DATA_TYPE = 'enum',
+						CONCAT(col.TABLE_NAME, '_', col.COLUMN_NAME),
+						col.DATA_TYPE)
+		) AS "dataType.Name",
+		IF (col.DATA_TYPE = 'enum', 'enum', 'base') AS "dataType.Kind",
+		col.COLUMN_TYPE LIKE '%unsigned%' AS "dataType.IsUnsigned"
+FROM
+		information_schema.columns AS col
+LEFT JOIN (
+		SELECT k.column_name, 1 AS IsPrimaryKey
 		FROM information_schema.table_constraints t
-			JOIN information_schema.key_column_usage k USING(constraint_name,table_schema,table_name)
-		WHERE table_schema = ? AND table_name = ? AND t.constraint_type='PRIMARY KEY' AND k.column_name = columns.column_name
-	)) AS "column.IsPrimaryKey",
-	IF (COLUMN_TYPE = 'tinyint(1)', 
-			'boolean', 
-			IF (DATA_TYPE='enum', 
-					CONCAT(TABLE_NAME, '_', COLUMN_NAME), 
-					DATA_TYPE)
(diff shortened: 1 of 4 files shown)


houten11:
Hmm, `4m9s` that's weird. I routinely run the generator with around 50 entities(tables and views) and it usually doesn't take more than a second.   
How many columns on average your tables/views have?

josephbuchma:
12 columns on average.
Another factor that makes it slow is that it's docker and I'm on mac with m1 chip (although I'm using [mysql 5.7 image build for arm](https://hub.docker.com/r/biarms/mysql))

josephbuchma:
LOL, just tried it on `mysql/mysql-server:latest` and it took 1s on master and 600ms with this changes. Maybe this is not a problem after all :D

josephbuchma:
Tried to run official mysql 5.7 image with x86 emulation in docker, and it's very slow. We're planning to upgrade mysql, but right now I need to be able to work with 5.7 locally, and I prefer running it in docker, so would appreciate getting this PR merged.

go-jet:
A bit more speed is always welcomed. :+1:

generator/mysql/query_set.go
```diff
@@ -27,39 +29,77 @@ ORDER BY table_name;
 		return nil, fmt.Errorf("failed to query %s metadata result: %w", tableType, err)
 	}
 
+	tblChan := make(chan int, len(tables))
+	errChan := make(chan error, 1)
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < runtime.NumCPU(); i++ {
```
go-jet, line 36: There is no need to limit the number of goroutines to number of CPUs.

generator/mysql/query_set.go
```diff
@@ -27,39 +29,77 @@ ORDER BY table_name;
...
+					case errChan <- fmt.Errorf("failed to get '%s' table columns metadata: %w", tables[tblIdx].Name, err1):
+						return
+					default:
+					}
+					return
+				}
+			}
+		}()
+	}
+
```
go-jet, line 54: Using `errgroup.Group` this code can be simplified and `tblChan` and `errChan` can be removed.

josephbuchma: Didn't want to add another dependency. If that's not a problem then sure, errgroup is awesome!

generator/mysql/query_set.go
```diff
@@ -27,39 +29,77 @@ ORDER BY table_name;
...
-SELECT COLUMN_NAME AS "column.Name", 
-	IS_NULLABLE = "YES" AS "column.IsNullable",
-	columns.COLUMN_COMMENT as "column.Comment",
-	(EXISTS(
-		SELECT 1
+SELECT
+		col.COLUMN_NAME AS "column.Name",
+		col.IS_NULLABLE = "YES" AS "column.IsNullable",
+		col.COLUMN_COMMENT AS "column.Comment",
+		pk.IsPrimaryKey AS "column.IsPrimaryKey",
```
go-jet, line 77: Looks good. Maybe `COALESCE` can be added around `pk.IsPrimaryKey` .

generator/mysql/query_set.go
go-jet: It's fine, because `golang.org/x/sync` is official golang repo.

generator/mysql/query_set.go
josephbuchma: Removed this and constrained it using 	`db.SetMaxOpenConns(32)` instead.

josephbuchma:
@go-jet I addressed all your comments, thanks for quick feedback!

generator/mysql/query_set.go
go-jet: Why is it needed?

go-jet: Aha, I see. We might exceed max number of connections on MySQL. Could you move it into [openConnection](https://github.com/go-jet/jet/blob/e2fe799f8ef4f529b8ca1a660c3b6a494a09dfc7/generator/mysql/mysql_generator.go#L85) so is not called multiple times.

go-jet:
LGTM. :+1:

PR #288, opened 2023-11-21
Add support for OF in row lock clauses

### LINKED ISSUE #285: Support `OF` in locking clause
I have a query doing LEFT JOIN across 2 tables modeled on a many-to-many relationship and locking rows to prevent concurrent updates.

My schema is as follows, minus some rows I trimmed out for simplification:

```sql
CREATE TABLE resources (
    id UUID PRIMARY KEY,
    name TEXT NOT NULL,
);

CREATE TABLE groups (
    id TEXT PRIMARY KEY,
    name TEXT NOT NULL
);

CREATE TABLE resources_groups (
    fk_resources_id UUID REFERENCES resources ON DELETE RESTRICT,
    fk_groups_id TEXT REFERENCES groups ON DELETE RESTRICT,
    PRIMARY KEY (fk_resources_id, fk_groups_id)
);
```

In essence I can have multiple resources and each resource can have multiple groups associated with it.

I then have an update API that executes a SQL query as follows:

```sql
SELECT
  resources.id AS "resources.id",
  resources.name AS "resources.name",
  groups.id AS "groups.id",
  groups.name AS "groups.name"
FROM public.resources
LEFT JOIN public.resources_groups ON (resources.id = resources_groups.fk_resources_id)
LEFT JOIN public.groups ON (groups.id = resources_groups.fk_groups_id)
WHERE (resources.id = 'd38f7d10-d265-4c10-b0c1-464d3af19d2b')
FOR UPDATE NOWAIT;
```

This produces an error:

> ERROR: FOR UPDATE cannot be applied to the nullable side of an outer join (SQLSTATE 0A000)

This in itself seems to be expected per https://www.postgresql.org/message-id/21634.1160151923@sss.pgh.pa.us. And it's likely I need to rethink my use of joins in general.

On the plus side, I found that adjusting the locking clause to:

```sql
FOR UPDATE OF resources NOWAIT;
```

Allows the query to work. However, this does not seem to be supported in jet, potentially becaise this seems to be a PostgeSQL specific feature.

```go
var m struct {
  Resource model.Resources
  Groups   []model.Groups
}

err := table.Resources.
  SELECT(table.Resources.AllColumns, table.Groups.AllColumns).
  FROM(
    table.Resources.
      LEFT_JOIN(table.ResourcesGroups, table.Resources.ID.EQ(table.ResourcesGroups.FkResourcesID)).
      LEFT_JOIN(table.Groups, table.Groups.ID.EQ(table.ResourcesGroups.FkGroupsID)),
  ).
  WHERE(table.Resources.ID.EQ(postgres.UUID(resourceID))).
  FOR(postgres.UPDATE().NOWAIT()). // No method named `OF` is available here
  QueryContext(ctx, conn, &m)
```

Support the use of `OF` in `postgres.RowLock` to allow outer joins to use locks.

In testing the revised query, I found that the table in the `OF` clause had to be just `resources` and the use of `public.resources` as used in the `FROM` clause did not work.

mattdowdell:
This adds support for statements such as `SELECT ... FOR UPDATE OF table NOWAIT` where `OF table` could not be specified previously. Fixes #285.

### DIFF
--- a/internal/jet/select_lock.go
+++ b/internal/jet/select_lock.go
@@ -4,12 +4,14 @@ package jet
 type RowLock interface {
 	Serializer
 
+	OF(...Table) RowLock
 	NOWAIT() RowLock
 	SKIP_LOCKED() RowLock
 }
 
 type selectLockImpl struct {
 	lockStrength       string
+	of                 []Table
 	noWait, skipLocked bool
 }
 
@@ -20,8 +22,13 @@ func NewRowLock(name string) func() RowLock {
 	}
 }
 
-func newSelectLock(lockStrength string) RowLock {
+func newSelectLock(lockStrength string) *selectLockImpl {
 	return &selectLockImpl{lockStrength: lockStrength}
+}
+
+func (s *selectLockImpl) OF(tables ...Table) RowLock {
+	s.of = tables
+	return s
 }
 
 func (s *selectLockImpl) NOWAIT() RowLock {
@@ -36,6 +43,23 @@ func (s *selectLockImpl) SKIP_LOCKED() RowLock {
 
 func (s *selectLockImpl) serialize(statement StatementType, out *SQLBuilder, options ...SerializeOption) {
 	out.WriteString(s.lockStrength)
+
+	if len(s.of) > 0 {
+		out.WriteString("OF")
+
+		for i, of := range s.of {
+			if i > 0 {
+				out.WriteString(", ")
+			}
+
+			table := of.Alias()
+			if table == "" {
+				table = of.TableName()
+			}
+
+			out.WriteIdentifier(table)
+		}
+	}
 
 	if s.noWait {
 		out.WriteString("NOWAIT")
--- a/tests/postgres/select_test.go
+++ b/tests/postgres/select_test.go
@@ -2251,6 +2251,18 @@ FOR`
 	}
 }
 
+func TestRowLockWithJoins(t *testing.T) {
+	query := SELECT(STAR).
+		FROM(
+			Film.
+				INNER_JOIN(FilmCategory, FilmCategory.FilmID.EQ(Film.FilmID)).
+				LEFT_JOIN(FilmActor, FilmActor.FilmID.EQ(Film.FilmID))).
+		LIMIT(1).
+		FOR(UPDATE().OF(Film, FilmCategory).NOWAIT())
+
+	testutils.AssertExecAndRollback(t, query, db, 1)
+}
+
 func TestQuickStart(t *testing.T) {
 
 	var expectedSQL = `
--- a/mysql/select_statement_test.go
+++ b/mysql/select_statement_test.go
@@ -123,6 +123,11 @@ SELECT table1.col_bool AS "table1.col_bool"
 FROM db.table1
 FOR SHARE NOWAIT;
 `)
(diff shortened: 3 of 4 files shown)


internal/jet/select_lock.go
```diff
@@ -37,6 +44,23 @@ func (s *selectLockImpl) SKIP_LOCKED() RowLock {
...
+			if i > 0 {
+				out.WriteString(",")
+			}
+
+			table := of.Alias()
+			if table == "" {
+				table = of.TableName()
+			}
+
+			out.WriteIdentifier(table, true)
```
go-jet, line 60: Why `alwaysQuote` table name?

go-jet:
Could you also add one [integration test](https://github.com/go-jet/jet/blob/master/tests/postgres/select_test.go) just to make sure everything works with live db.

internal/jet/select_lock.go
mattdowdell: I think this was left over from me experimenting with what the expected behaviour was. For some reason, the tests always ended up quoting regardless of what this value was.

Will update to false as it doesn't seem always necessary.

```suggestion
			out.WriteIdentifier(table, false)
```

internal/jet/select_lock.go
```diff
@@ -37,6 +44,23 @@ func (s *selectLockImpl) SKIP_LOCKED() RowLock {
...
+			if i > 0 {
+				out.WriteString(", ")
+			}
+
+			table := of.Alias()
+			if table == "" {
+				table = of.TableName()
+			}
+
+			out.WriteIdentifier(table, false)
```
go-jet, line 60: You'll need to remove `alwaysQuote` completely to avoid the quote. It is an optional parameter.

tests/postgres/select_test.go
```diff
@@ -2251,6 +2251,37 @@ FOR`
...
+func TestRowLockWithJoins(t *testing.T) {
+	expectedSQL := `
+SELECT *
+FROM dvds.film
+     INNER JOIN dvds.film_category ON (film_category.film_id = film.film_id)
+     LEFT JOIN dvds.film_actor ON (film_actor.film_id = film.film_id)
+LIMIT 1
+FOR UPDATE OF "film", "film_category" NOWAIT;
+`
+	query := Film.
```
go-jet, line 2263: It doesn't make sense to use both classical and jet select statement in the same query. Since there is a `FROM` clause `Film.` can be removed.

tests/postgres/select_test.go
```diff
@@ -2251,6 +2251,37 @@ FOR`
...
+
+	tx, _ := db.Begin()
+
+	res, err := query.Exec(tx)
+	require.NoError(t, err)
+	rowsAffected, _ := res.RowsAffected()
+	require.Equal(t, rowsAffected, int64(1))
+
+	err = tx.Rollback()
+	require.NoError(t, err)
```
go-jet, line 2282: 2274-2282, can be replaced with a `testutils.AssertExecAndRollback(t, query, db, 1)`

internal/jet/select_lock.go
mattdowdell: Done in https://github.com/go-jet/jet/pull/288/commits/92a923b98be3182bcaa2ecd8d229985cf76baf93

go-jet:
LGTM. :+1:

PR #307, opened 2024-01-31
Feature/daterange

sarkan-ag5:
This pull request introduces support for PostgreSQL range type fields. With this enhancement, Jet gains the ability to handle PostgreSQL range types seamlessly, enabling users to define and manipulate range-based data directly. The implementation is based on PostgreSQL [documents](https://www.postgresql.org/docs/current/rangetypes.html).


internal/jet/range_expression.go
```diff
@@ -0,0 +1,90 @@
...
+
+	CONTAINS(rhs Expression) BoolExpression
+	OVERLAP(rhs RangeExpression) BoolExpression
+	UNION(rhs RangeExpression) RangeExpression
+	INTERSECTION(rhs RangeExpression) RangeExpression
+	DIFFERENCE(rhs RangeExpression) RangeExpression
+}
+
+type rangeInterfaceImpl struct {
+	parent RangeExpression
```
go-jet, line 23: Can be just `parent Expression`

internal/jet/operators.go
```diff
@@ -69,6 +69,16 @@ func GtEq(lhs, rhs Expression) BoolExpression {
...
+// Contains returns a representation of "a @> b"
+func Contains(lhs RangeExpression, rhs Expression) BoolExpression {
+	return newBinaryBoolOperatorExpression(lhs, rhs, "@>")
+}
+
+// Overlap returns a representation of "a && b"
+func Overlap(lhs, rhs RangeExpression) BoolExpression {
+	return newBinaryBoolOperatorExpression(lhs, rhs, "&&")
+}
+
```
go-jet, line 81: If these two methods are called only once from `range_expression.go`, I think we can go without them.

postgres/cast.go
```diff
@@ -43,6 +43,8 @@ type cast interface {
 	AS_TIMESTAMPZ() TimestampzExpression
 	// Cast expression AS interval type
 	AS_INTERVAL() IntervalExpression
+	// Cast expression AS range type
+	AS_DATERANGE() RangeExpression
```
go-jet, line 47: Casts for other range types.

internal/jet/literal_expression.go
```diff
@@ -326,6 +326,46 @@ func formatNanoseconds(nanoseconds ...time.Duration) string {
...
+	rangeLiteral.rangeInterfaceImpl.parent = &rangeLiteral
+
+	return &rangeLiteral
+}
+
+func DateRange(lowIncluded, highIncluded bool, lowDate, highDate time.Time) RangeExpression {
+	return RangeRaw(getRangeFromDate(lowIncluded, highIncluded, lowDate, highDate))
+}
+
+func getRangeFromDate(lowIncluded, highIncluded bool, lowDate, highDate time.Time) string {
```
go-jet, line 347: Postgres alrady defines how ranges are [constructed](https://www.postgresql.org/docs/16/rangetypes.html#RANGETYPES-CONSTRUCT). There is a third string parameter that defines if lower/upper bound is included.
```sql
SELECT numrange(1.0, 14.0, '(]');
```
To be consistent I think we should use the same convention.

go-jet:
Hi @sarkan-ag5,

Nice work overall.

Couple of additional comments:
 - generator would need to be updated next with a new sql builder types, check - https://github.com/go-jet/jet/blob/23cb5dcfbc0aac3bf36c78847133cd768a5c8751/generator/template/sql_builder_template.go#L144
   I've made some changes in generator recently, so you might need first to rebase with master.
- for integration tests we can add one more table `range_table` to `test_sample` database. Check [test folder](https://github.com/go-jet/jet/tree/master/tests) how to setup and run intergration tests.

internal/jet/literal_expression.go
```diff
@@ -326,6 +326,46 @@ func formatNanoseconds(nanoseconds ...time.Duration) string {
 	return ""
 }
 
+// ---------------------------------------------------//
+type rangeLiteral struct {
+	rangeInterfaceImpl
+	literalExpressionImpl
+}
+
+func RangeRaw(value string) RangeExpression {
```
go-jet, line 335: This should be:

```golang
func RawRange(raw string, namedArgs ...map[string]interface{}) RangeExpression {
	return RangeExp(Raw(raw, namedArgs...))
}
```

internal/jet/range_expression.go
```diff
@@ -0,0 +1,90 @@
+package jet
+
+// RangeExpression is interface for date range types
+type RangeExpression interface {
```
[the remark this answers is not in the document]
sarkan-ag5: To be honest, the reason I didn't use generics in that part of the code was because your coding style didn't include any generic functions. However, after receiving your comment, I will definitely add generics to that part because it's the best place to use them

internal/jet/operators.go
sarkan-ag5: All of operators are in that file maybe its better to keep these two here as well.

internal/jet/range_expression.go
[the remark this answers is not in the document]
go-jet: > And the main change will be golang version, which right now its 1.11 and has to be updated to 1.18 at least. Are you sure about it?
Yeah, that's fine.

internal/jet/literal_expression.go
```diff
@@ -326,6 +326,102 @@ func formatNanoseconds(nanoseconds ...time.Duration) string {
 	return ""
 }
 
+// ---------------------------------------------------//
+type rangeLiteral struct {
```
go-jet, line 330: `rangeLiteral` can now be removed.

internal/jet/literal_expression.go
```diff
@@ -326,6 +326,102 @@ func formatNanoseconds(nanoseconds ...time.Duration) string {
...
+	rangeInterfaceImpl
+	literalExpressionImpl
+}
+
+func RawRange(raw string, namedArgs ...map[string]interface{}) RangeExpression {
+	return RangeExp(Raw(raw, namedArgs...))
+}
+
+func NumRange(lowNum, highNum *int, bounds string) RangeExpression {
```
go-jet, line 339: - `NumRange` should accept `NumericExpression`, since postgres does allow expression to appear as a parameter. For instance:   
```sql
select numrange(MIN(actor.actor_id), 100)
from dvds.actor;
```
- To avoid dealing with pointers and to represent infinities we need to define a new type `PlusInfinity` and `MinusInfinity`, similary to existing [`NULL` type](https://github.com/go-jet/jet/blob/e03773a79ec5e78fdc2ddd9ceb8057dd0d58c0c6/internal/jet/literal_expression.go#L333). The type would be `Expression`, but we can then convert it to `IntegerExpressoin` or `FloatExpression`, with `IntExp(PlusInfinity)`, `FloatExp(MinusInfinity)`
- according to postgres `bounds` can be omitted, we can simulate that with variadic arguments `,bounds... string)`

internal/jet/literal_expression.go
```diff
@@ -326,6 +326,102 @@ func formatNanoseconds(nanoseconds ...time.Duration) string {
...
+
+	low := ""
+	if lowNum != nil {
+		low = fmt.Sprint(*lowNum)
+	}
+	high := ""
+	if highNum != nil {
+		high = fmt.Sprint(*highNum)
+	}
+	return RawRange(fmt.Sprintf("%clowerNum,higherNum%c", bounds[0], bounds[1]), map[string]interface{}{
```
go-jet, line 350: `NumRange` should be a [constructor function]( https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-CONSTRUCT), not a literal.
`numrange(1.0, 14.0, '(]');` vs `'[3,7)'::int4range`.

You can use function helper function:

```golang
return RangeExp(NewFunc("numrange", allparameters, nil))
```

postgres/cast.go
```diff
@@ -159,3 +167,23 @@ func (b *castImpl) AS_TIMESTAMPZ() TimestampzExpression {
 func (b *castImpl) AS_INTERVAL() IntervalExpression {
 	return IntervalExp(b.AS("interval"))
 }
+
+// Cast expression AS range type
```
go-jet, line 171: What about `int4range` and `int8range`?

postgres/literal.go
```diff
@@ -138,3 +138,23 @@ func Timestampz(year int, month time.Month, day, hour, minute, second int, milli
 func TimestampzT(t time.Time) TimestampzExpression {
 	return CAST(jet.TimestampzT(t)).AS_TIMESTAMPZ()
 }
+
+// DateRange creates new date range literal expression from two dates
+func DateRange(lowDate, highDate time.Time, bounds string) RangeExpression {
+	return CAST(jet.DateRange(lowDate, highDate, bounds)).AS_DATERANGE()
```
go-jet, line 144: If `DateRange` is function constructor then additional cast is probably unnecessary. Because generated sql would be `daterange(start,end)::daterange`

internal/jet/range_expression.go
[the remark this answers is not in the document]
sarkan-ag5: Could you take a look at current state, please? and If it is possible provide more detail about this part, I like to know about your idea

internal/jet/func_expression.go
```diff
@@ -843,3 +855,60 @@ func newTimestampzFunc(name string, expressions ...Expression) *timestampzFunc {
...
+	rangeFunc.rangeInterfaceImpl.parent = rangeFunc
+
+	return rangeFunc
+}
+
+func NumRange(lowNum, highNum NumericExpression, bounds ...StringExpression) RangeExpression {
+	if len(bounds) != 2 {
+		panic("bound must contains 2 elements for lower and higher bound")
+	}
+	return RangeExp(NewFunc("numrange", []Expression{lowNum, highNum, bounds[0], bounds[1]}, nil))
```
go-jet, line 878: I don't think this will work. 3rd parameter is single string expression. So you just need to pass `,bounds...`.   
The only condition for bounds length is that it can't be greater then 1.

internal/jet/range_expression.go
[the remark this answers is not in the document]
go-jet: Current state looks good. Now you just need to add a template to `RangeExpression`. Something like:

```golang
type Range[T Expression] interface {
	Expression

	EQ(rhs Range[T]) BoolExpression
        ....
	CONTAINS(rhs T) BoolExpression
	OVERLAP(rhs Range[T]) BoolExpression
        ...
}
```
After this, you'll probably have to add the same template to each of the new range functions and structs.
And eventually in `postgres` package you can export new types:

```golang
// new expressions
type Int4RangeExpression = jet.Range[IntegerExpression]
type Int8RangeExpression = jet.Range[IntegerExpression]
// ....
// new columns
var Int4RangeColumn = jet.Range[IntegerExpression]
type ColumnInt4Range = jet.ColumnRange[IntegerExpression]
// ...
// etc...
```

internal/jet/func_expression.go
```diff
@@ -468,6 +468,18 @@ func REGEXP_LIKE(stringExp StringExpression, pattern StringExpression, matchType
 	return newBoolFunc("REGEXP_LIKE", stringExp, pattern)
 }
 
+//----------Range Type Functions ----------------------//
+
+// LOWER_BOUND returns range expressions lower bound
+func LOWER_BOUND[T Expression](rangeExpression Range[T]) Range[T] {
```
go-jet, line 474: `LOWER_BOUND` and `UPPER_BOUND` should return range element type T.

internal/jet/func_expression.go
```diff
@@ -843,3 +855,42 @@ func newTimestampzFunc(name string, expressions ...Expression) *timestampzFunc {
 func Func(name string, expressions ...Expression) Expression {
 	return NewFunc(name, expressions, nil)
 }
+
+type rangeFunc[T Expression] struct {
```
go-jet, line 859: I don't think `rangeFunc[T Expression]` is needed, because you can do always do `RangeExp[T](NewFunc(....))`

internal/jet/func_expression.go
```diff
@@ -843,3 +855,42 @@ func newTimestampzFunc(name string, expressions ...Expression) *timestampzFunc {
...
+func NewRangeFunc[T Expression](name string, expressions ...Expression) *rangeFunc[T] {
+	rangeFunc := &rangeFunc[T]{}
+
+	rangeFunc.funcExpressionImpl = *NewFunc(name, expressions, rangeFunc)
+	rangeFunc.rangeInterfaceImpl.parent = rangeFunc
+
+	return rangeFunc
+}
+
+func NumRange(lowNum, highNum NumericExpression, bounds StringExpression) Range[NumericExpression] {
```
go-jet, line 874: `bounds` is optional, so `bounds ...StringExpression`

postgres/expressions.go
```diff
@@ -99,6 +124,7 @@ var (
 	RawTimestamp  = jet.RawTimestamp
 	RawTimestampz = jet.RawTimestampz
 	RawDate       = jet.RawDate
+	// TODO: add raw range
```
go-jet, line 127: // TODO: add raw range

go-jet:
Looks, good. The next steps would be to [update generator and add integration tests](https://github.com/go-jet/jet/pull/307#issuecomment-1923541349).

internal/jet/func_expression.go
```diff
@@ -468,6 +468,33 @@ func REGEXP_LIKE(stringExp StringExpression, pattern StringExpression, matchType
...
+}
+
+// UPPER_BOUND returns range expressions upper bound
+func UPPER_BOUND[T Expression](rangeExpression Range[T]) T {
+	return rangeTypeCaster[T](rangeExpression, NewFunc("UPPER", []Expression{rangeExpression}, nil))
+}
+
+func rangeTypeCaster[T Expression](rangeExpression Range[T], exp Expression) T {
+	var i Expression
+	switch rangeExpression.(type) {
```
go-jet, line 485: I don't think this will compile. Go does not allow using generics and type switch in the same function.
I also don't see how we can return `T` in this case, so I think we should just return `Expression`. The user will have to, if needed, add type manually , like: `IntExp(UPPER_BOUND(table.Range)).ADD()`

internal/jet/func_expression.go
```diff
@@ -843,3 +870,35 @@ func newTimestampzFunc(name string, expressions ...Expression) *timestampzFunc {
...
+func DateRange(lowTs, highTs DateExpression, bounds ...StringExpression) Range[DateExpression] {
+	return RangeExp[DateExpression](NewFunc("daterange", rangeFuncParamCombiner[DateExpression](lowTs, highTs, bounds...), nil))
+}
+
+func rangeFuncParamCombiner[T Expression](low, high T, bounds ...StringExpression) []Expression {
+	exp := []Expression{low, high}
+	if len(bounds) != 0 {
+		exp = append(exp, bounds[0])
+	}
+	return exp
```
go-jet, line 903: It can be a bit simpler: `append([]Expression{low, high}, bounds...)`

sarkan-ag5: we can not use this syntax because the bounds type is StringExpression
Error:
`Cannot use 'bounds' (type []StringExpression) as the type Expression Type does not implement 'Expression' as some methods are missing: serialize(statement StatementType, out *SQLBuilder, options ...SerializeOption) serializeForProjection(statement StatementType, out *SQLBuilder) fromImpl(subQuery SelectTable) Projection`

internal/jet/func_expression.go
sarkan-ag5: As you can see I used switch on first parameter not `T` itself and also here `internal/jet/func_expression_test.go:206` just tested with a simple IntegerExpression and there isn't any problem.
Returning an expression make it more simple and less type safe, but I think as you can see we can return T itself with that switch case so why not?

go-jet: Hmm. If it works it is great, I wasn't aware it was possible.

internal/jet/func_expression.go
go-jet: Yeah, you right.

tests/postgres/range_test.go
```diff
@@ -0,0 +1,419 @@
...
+     sample_ranges.timestamp_range AS "sample_ranges.timestamp_range",
+     sample_ranges.timestampz_range AS "sample_ranges.timestampz_range",
+     sample_ranges.int4_range AS "sample_ranges.int4_range",
+     sample_ranges.int8_range AS "sample_ranges.int8_range",
+     sample_ranges.num_range AS "sample_ranges.num_range"
+FROM test_sample.sample_ranges
+WHERE sample_ranges.date_range @> '2023-12-12'::date;
+`
+
+	query := SELECT(SampleRanges.AllColumns).
```
go-jet, line 31: Could you include also here calls to UPPER_BOUND, LOWER_BOUND, as well as range constructors and infinities?

go-jet:
I think we are close now to merge this PR. Could you squash all the commits into one. I'm not completely sure but if I do squash and merge your contribution might be deleted from git history.

go-jet:
Looks great. :+1:

PR #309, opened 2024-02-03
Improve performance of mysql generator

jupp0r:
This change improves performance for generating mysql models for databases with large number of tables. In my local testing for a database with about 1000 tables and 140k columns, generation time was reduced from about 1h to less than one second.


go-jet:
Hi @jupp0r,

Whats your setup? Are you running MySQL 5.7 on a docker? Mac silicon?

> generation time was reduced from about 1h to less than one second.

One minut or one second?

generator/mysql/query_set.go
```diff
@@ -28,20 +32,45 @@ ORDER BY table_name;
...
+
+	for _, column := range columns {
+		tableIndex, ok := tableIndexLookup[column.TableName]
+		if !ok {
+			return nil, fmt.Errorf("table %s not found in the metadata for column %v+", column.TableName, column)
+		}
+
+		table := &tables[tableIndex]
+		table.Columns = append(table.Columns, column.Column)
 	}
```
go-jet, line 59: I think, to solve the N+1 problem, a better approach would be to use jet `qrm` directly. This means merging everything in one query and letting `qrm.Query` group the result. You would also need to add `sql:"primary_key"` tag to `Name` field of `metadata.Table` struct so `qrm` will work.

jupp0r:
> Hi @jupp0r,
> 
> Whats your setup? Are you running MySQL 5.7 on a docker? Mac silicon?
> 
> > generation time was reduced from about 1h to less than one second.
> 
> One minut or one second?

One second. This is on an Apple M1 Max Macbook Pro running Percona 5.7 natively (no docker). Each of the queries was taking more than 30s for some reason.

generator/mysql/query_set.go
jupp0r: You were right, this is much nicer now.

generator/mysql/query_set.go
```diff
@@ -56,30 +29,33 @@ SELECT
 		) AS "dataType.Name",
 		IF (col.DATA_TYPE = 'enum', 'enum', 'base') AS "dataType.Kind",
 		col.COLUMN_TYPE LIKE '%unsigned%' AS "dataType.IsUnsigned"
-FROM
+FROM INFORMATION_SCHEMA.tables AS t
+LEFT OUTER JOIN
```
go-jet, line 33: It can be INNER JOIN. It is not possible to have a table without columns in MySQL.

generator/mysql/query_set.go
```diff
@@ -56,30 +29,33 @@ SELECT
...
-			AND t.table_name = ?
 			AND t.constraint_type = 'PRIMARY KEY'
-) AS pk ON col.COLUMN_NAME = pk.column_name
-WHERE
-		col.table_schema = ?
-		AND col.table_name = ?
+) AS pk ON col.COLUMN_NAME = pk.column_name AND col.table_name = pk.table_name 
+WHERE t.table_schema = ?
+	AND t.table_type = ?
+	AND col.table_schema = ?
```
go-jet, line 45: `AND col.table_schema = ?` seems redundant.

go-jet:
Looks good. :+1:

PR #311, opened 2024-02-06
Handle plus and minus identifiers better

jupp0r:
This change improves handling of db identifiers:
- it handles "+" in identifiers
- it improves handling of identifiers that end with plus or minus like "Entity+" and "Entity-" as enum values that would otherwise produce conflicting values

### DIFF
--- a/internal/utils/dbidentifier/dbidentifier_test.go
+++ b/internal/utils/dbidentifier/dbidentifier_test.go
@@ -1,8 +1,9 @@
 package dbidentifier
 
 import (
-	"github.com/stretchr/testify/require"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestToGoIdentifier(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/utils/dbidentifier/dbidentifier_test.go
```diff
@@ -22,4 +23,7 @@ func TestToGoIdentifier(t *testing.T) {
 	require.Equal(t, ToGoIdentifier("My_Table"), "MyTable")
 	require.Equal(t, ToGoIdentifier("My Table"), "MyTable")
 	require.Equal(t, ToGoIdentifier("My-Table"), "MyTable")
+
+	require.Equal(t, ToGoIdentifier("My-Table+"), "MyTablePlus")
+	require.Equal(t, ToGoIdentifier("My-Table-"), "MyTableMinus")
```
go-jet, line 28: It feels like quick fix for your use case. But there are many more. The invalid character can also appear at the start or at the middle of identifier. From the perspective of this library, jet will replace all go identifier invalid characters with `_`(There is an improvement for this at branch [bug290](https://github.com/go-jet/jet/commit/48a3e78dc714414e8910e1c57f7febab1364917c), yet to be merged to master).   
And then if there is a conflict developer can use [generator customization](https://github.com/go-jet/jet/wiki/Generator#generator-customization) to replace conflicting name with a name of his choosing.

jupp0r: Makes sense, I'll workaround using generator customization for now.

go-jet:
I've changed my mind on this one. There were couple of bugs opened before for similar invalid character issues.  To prevent any additional bugs opened, I've made a [change](https://github.com/go-jet/jet/commit/6b098b8e41fc8adc22418202019b1e2cb2441b58) that replaces all invalid ascii characters with they description.   
In your case, you would still need to customize generator because minus and hyphen are the same character, and hyphen is first replaced with underscore, and underscores at the end of identifier are removed.

PR #325, opened 2024-02-21
mysql: added a helper to compare UUID strings with uuid_to_bin

### LINKED ISSUE #321: Customize Generated model struct tags
  Its pretty annoying when I generate models and I cant customize their struct tags , specifically I'd beable to use the generated models if it had json struct tags to make the Fields follow json convention but as they cannot i have to make custom models for each one that i will need to serialize out to JSON

I'd like something where you can just tell the generator to add json struct tags to the generated model.

### LINKED ISSUE #324: MySQL: A utility fuinction for comparision of UUID fields
Assume working with a table that has UUIDs stored in a `varbinary` field. Here's an example:

```sql
CREATE TABLE `some_table` (
  `id` varbinary(16) NOT NULL,
  `some_column` varchar(50) DEFAULT NULL,
  PRIMARY KEY (`id`)
);
```

Querying data by `id` in this table would look like this:

```
select * from `some_table` where id = uuid_to_bin('0000-0000...');
```

The type of the `id` column would be of `mysql.ColumnString` in the generated code.

```go
type someTableTable struct {
    mysql.Table

    ID         mysql.ColumnString
    SomeColumn mysql.ColumnString
}
```

Generating a query using the `.EQ` method on `SomeTable.ID` would result in a query that looks like: `select .... where `id` = '0000-0000-.....';` `mysql.UUID` also appears to be just a wrapper that takes a `fmt.Stringer` and generates the same output as the `.EQ` method.

To get the desired output, we can use something like this:

```go
id := "00000-00000-....."
expr := mysql.Func("uuid_to_bin", mysql.String(id))

statement := table.SomeTable.SELECT(table.SomeTable.AllColumns).
        WHERE(table.SomeTable.ID.EQ(mysql.StringExp(expr)))
```

Would it make sense for the library to add a couple of utility functions to shorten this? For example, `functions.go` could have these:

```go
func UUIDToBin(str fmt.Sringer) StringExpression { ... }

func StringUUIDToBin(str string) StringExpression { ... }
```

With this, the user code can be shortened as:
```go
id := "00000-00000-....."
statement := table.SomeTable.SELECT(table.SomeTable.AllColumns).
        WHERE(table.SomeTable.ID.EQ(mysql.StringUUIDToBin(id)))
```

realbucksavage:
Adds a helper function to resolve #324 .

### DIFF
--- a/mysql/literal.go
+++ b/mysql/literal.go
@@ -1,7 +1,6 @@
 package mysql
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/go-jet/jet/v2/internal/jet"
@@ -56,17 +55,6 @@ var String = jet.String
 // UUID is a helper function to create string literal expression from uuid object
 // value can be any uuid type with a String method
 var UUID = jet.UUID
-
-// UUIDToBin takes ay object with a String method and calls StringUUIDToBin.
-func UUIDToBin(str fmt.Stringer) StringExpression {
-	return StringUUIDToBin(str.String())
-}
-
-// StringUUIDToBin is a helper function that calls "uuid_to_bin" function on the passed value.
-func StringUUIDToBin(str string) StringExpression {
-	fn := Func("uuid_to_bin", String(str))
-	return StringExp(fn)
-}
 
 // Date creates new date literal
 func Date(year int, month time.Month, day int) DateExpression {
--- a/mysql/literal.go
+++ b/mysql/literal.go
@@ -1,8 +1,10 @@
 package mysql
 
 import (
-	"github.com/go-jet/jet/v2/internal/jet"
+	"fmt"
 	"time"
+
+	"github.com/go-jet/jet/v2/internal/jet"
 )
 
 // Keywords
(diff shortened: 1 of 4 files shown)


mysql/literal.go
```diff
@@ -55,6 +57,17 @@ var String = jet.String
 // value can be any uuid type with a String method
 var UUID = jet.UUID
 
+// UUIDToBin takes ay object with a String method and calls StringUUIDToBin.
+func UUIDToBin(str fmt.Stringer) StringExpression {
```
go-jet, line 61: * Naming convention is to use all uppercase for all sql functions, so the name should be `UUID_TO_BIN`
* `uuid_to_bin` accepts not only parameters, but also columns and expression, so function parameter should be `StringExpression`
* `functions.go` is probably more appropriate place.

realbucksavage: @go-jet makes sense. This also eliminates the need for the `StringUUIDToBin` function.

PR #335, opened 2024-03-19
ON CONFLICT DO NOTHING without a conflict target is valid SQL

BillBuilt:
ON CONFLICT DO NOTHING without a conflict target is valid SQL according to Postgres.

https://www.postgresql.org/docs/15/sql-insert.html#SQL-ON-CONFLICT

### DIFF
--- a/postgres/clause.go
+++ b/postgres/clause.go
@@ -44,7 +44,7 @@ func (o *onConflictClause) DO_UPDATE(action conflictAction) InsertStatement {
 }
 
 func (o *onConflictClause) Serialize(statementType jet.StatementType, out *jet.SQLBuilder, options ...jet.SerializeOption) {
-	if len(o.indexExpressions) == 0 && o.constraint == "" {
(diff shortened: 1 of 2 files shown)


go-jet:
Hi @BillBuilt, good catch.

postgres/clause.go
```diff
@@ -44,7 +44,7 @@ func (o *onConflictClause) DO_UPDATE(action conflictAction) InsertStatement {
 }
 
 func (o *onConflictClause) Serialize(statementType jet.StatementType, out *jet.SQLBuilder, options ...jet.SerializeOption) {
-	if len(o.indexExpressions) == 0 && o.constraint == "" {
+	if len(o.indexExpressions) == 0 && o.constraint == "" && o.do != jet.Keyword("DO NOTHING") {
```
go-jet, line 47: Why not `o.do != nil`?

BillBuilt: Yea That's probably better.

Issue #355, opened 2024-06-10
Support for keyword arguments to functions


kblomster:
### Problem
Both standard SQL and many dialects have function signatures that look something like `SOME_FUNC(arg1, arg2 KEYWORD arg3)`. To name a few specific examples, we have things like:

* Standard SQL `SUBSTRING(extraction_string FROM starting_position [FOR length] [COLLATE collation_name])`
* MySQL `TRIM([{BOTH | LEADING | TRAILING} [remstr] FROM] str)`
* Postgres and SQLite ordered aggregations, such as (pg) `ARRAY_AGG(a ORDER BY b DESC)`
* Postgres nonsense like `json_objectagg ( [ { key_expression { VALUE | ':' } value_expression } ] [ { NULL | ABSENT } ON NULL ] [ { WITH | WITHOUT } UNIQUE [ KEYS ] ] [ RETURNING data_type [ FORMAT JSON [ ENCODING UTF8 ] ] ])`

In Jet, currently all of these require using a `Raw` expression, which is inconvenient. This is especially true if the expressions you want to use as arguments are anything other than column names - column names are easily accessible in a type-safe way, but there's no public API for converting an arbitrary expression to a SQL fragment string.

### Possible solutions
Most convenient for me as a user would be if Jet knew the signature of every builtin function in every dialect and had type-safe builders for every one. A very rough sketch of what an API like that could look like:

```golang
// first thing that comes to mind: ARRAY_AGG would take an AggregateArgs object,
// that knows how to separate the aggregated expressions from the ORDER BY clause
SELECT(
    ARRAY_AGG(
        AggregateArgs(a, b).
            ORDER_BY(c.DESC())))
```

That's probably unreasonable though, so the next best thing would be to have some kind of support for this in `jet.FUNC`, or a variant of it that supported things like this. Currently you can't just pass the keywords as raw arguments because they get commas inserted between them, which is not the syntax we need.

I'm kind of stumped at what the signature of a `jet.FUNC` that does what I want would look like, though. I'm imagining something like

```golang
jet.FUNC("ARRAY_AGG", SomeCol, jet.Raw("ORDER BY"), OtherCol.DESC())
```

but how do we tell it there shouldn't be a comma before `ORDER BY`?

Alternately, maybe this could be addressed in an entirely different manner by exposing a public `ToSQL` method on expressions so we could more easily build raw expressions from component parts, but that has an entirely different set of API design headaches.

go-jet:
[...]

PR #361, opened 2024-07-11
feat: export CustomExpression and Token for dialects to use 

mathieuk:
Hey Goran, nice lib! I hope you'll consider allowing a change like this. Doing this allows a particular SQL dialect to implement additional operators/expressions that may deviate from default SQL.

Background: I'm playing with a dialect that supports this notation:

  SELECT * FROM table WHERE any_match ( x -> x IS NOT NULL, someField )

That'd be something I'd want to implement just for that dialect (in its own dialect package in jet) but for that I need access to the serializer method. 

This pull request has no functional changes, just a refactoring to expose `jet.newCustomExpression` to other jet packages.

### DIFF
--- a/internal/jet/expression.go
+++ b/internal/jet/expression.go
@@ -278,7 +278,7 @@ type customExpression struct {
 	parts []Serializer
 }
 
-func newCustomExpression(parts ...Serializer) Expression {
+func NewCustomExpression(parts ...Serializer) Expression {
 	ret := customExpression{
 		parts: parts,
 	}
--- a/internal/jet/expression.go
+++ b/internal/jet/expression.go
@@ -278,7 +278,7 @@ type customExpression struct {
 	parts []Serializer
 }
 
-func NewCustomExpression(parts ...Serializer) Expression {
+func CustomExpression(parts ...Serializer) Expression {
 	ret := customExpression{
 		parts: parts,
 	}
--- a/internal/jet/func_expression.go
+++ b/internal/jet/func_expression.go
@@ -548,7 +548,7 @@ func TO_TIMESTAMP(timestampzStr, format StringExpression) TimestampzExpression {
 
 // EXTRACT extracts time component from time expression
 func EXTRACT(field string, from Expression) Expression {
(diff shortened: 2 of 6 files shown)


go-jet:
Hi @mathieuk ,
Check this [comment](https://github.com/go-jet/jet/issues/355#issuecomment-2160308646), I think it applies in your case as well. If [CustomExpression](https://github.com/go-jet/jet/blob/b13f6ab08d1005e5f6f2d64695d26d639e1dacd7/internal/jet/expression.go#L281) func and [Token](https://github.com/go-jet/jet/blob/b13f6ab08d1005e5f6f2d64695d26d639e1dacd7/internal/jet/serializer.go#L105) are exported you should be able to write something like:

```golang
// any_match ( x -> x IS NOT NULL, someField )
func ANY_MATCH(x Expression, cond BoolExpression, someField Expression) BoolExpression {
        return BoolExp(Func("any_match", CustomExpression(x, Token("->"), cond, b)))
}
// usage:
ANY_MATCH(x, x.IS_NOT_NULL, someField)
```

mathieuk:
@go-jet That seems to make thing much easier, great! I've updated the pull request accordingly :)

internal/jet/expression.go
```diff
@@ -278,7 +278,7 @@ type customExpression struct {
 	parts []Serializer
 }
 
-func newCustomExpression(parts ...Serializer) Expression {
+func NewCustomExpression(parts ...Serializer) Expression {
```
go-jet, line 281: Without `New`. Just `CustomExpression` is fine.

go-jet:
You'll also need to export `CustomExpression` and `Token` from internal package to each of dialect packages `postgres/expression.go`, etc...

internal/jet/expression.go
mathieuk: applied

mathieuk:
> You'll also need to export `CustomExpression` and `Token` from internal package to each of dialect packages `postgres/expression.go`, etc...

Done.

go-jet:
LGTM. :+1:

PR #377, opened 2024-08-28
Bug/postgres array detection

realrunner:
The current column metadata query fails to identify array data types correctly.


go-jet:
Thanks. LGMT. :+1:

PR #380, opened 2024-09-04
Adding support for arrays

arjen-ag5:
Hello,

I'm working on support for arrays in go-jet. I was wondering if this is a feature that is considered in the long run and if it maybe merged into upstream in the long term. I'd like to show the progress I made so far and if it's any use trying to get it merged at all. If not I will not spend any effort in making it work with the different database types. Currently I only implemented postgres.

I can see from the Range support that is added earlier that the types from the `pgtypes` packages are being used. I took this into consideration for use as the SQL types in the generated table types. But the `pgtypes.Array*` structs are tailored to represent the internal representation of the postgresql array types. In constrast to the `pq.Array*` types which are much more user friendly, thats why I chose to use the latter.


go-jet:
Hi @arjen-ag5, thanks for starting the work on arrays.

generator/template/model_template.go
```diff
@@ -268,6 +269,11 @@ func getGoType(column metadata.Column) interface{} {
 
 // toGoType returns model type for column info.
 func toGoType(column metadata.Column) interface{} {
+	// We don't support multi-dimensional arrays
+	if column.DataType.Dimensions > 1 {
+		return ""
+	}
+
```
go-jet, line 276: Multi dimensional arrays are just arrays containing other arrays. In our  case `Array[Array[StringExpression]]`. But it is fine with PR to go with single dimension arrays only.

generator/template/model_template.go
```diff
@@ -333,6 +339,14 @@ func toGoType(column metadata.Column) interface{} {
...
 	case "numrange":
 		return pgtype.Numrange{}
+	case "bool[]":
+		return pq.BoolArray{}
+	case "integer[]", "int4[]":
+		return pq.Int32Array{}
+	case "bigint[]":
+		return pq.Int64Array{}
+	case "text[]", "jsonb[]", "json[]":
+		return pq.StringArray{}
```
go-jet, line 349: I'm fine using `pq`  types for one dimensional arrays. Maybe `pgtype` can be used for multi dimensional arrays.
Other types as well, time, date, etc...

generator/template/sql_builder_template.go
```diff
@@ -145,11 +146,42 @@ func DefaultTableSQLBuilderColumn(columnMetaData metadata.Column) TableSQLBuilde
...
+		columnMetaData.DataType.Kind != metadata.RangeType &&
+		columnMetaData.DataType.Kind != metadata.ArrayType {
 		return "String"
 	}
 
-	switch strings.ToLower(columnMetaData.DataType.Name) {
+	typeName := columnMetaData.DataType.Name
+	columnName := columnMetaData.Name
+
+	if columnMetaData.DataType.Kind == metadata.ArrayType {
```
go-jet, line 157: Why not add a case switch for each of the array types, as it is already done for other types?

generator/template/sql_builder_template.go
```diff
@@ -190,8 +222,7 @@ func getSqlBuilderColumnType(columnMetaData metadata.Column) string {
 	case "numrange":
 		return "NumericRange"
 	default:
-		fmt.Println("- [SQL Builder] Unsupported sql column '" + columnMetaData.Name + " " + columnMetaData.DataType.Name + "', using StringColumn instead.")
```
go-jet, line 193: Even if we add array types, there still gonna be some unsupported types, so warning message can remain.

internal/jet/array_expression.go
```diff
@@ -0,0 +1,93 @@
+package jet
+
+// ArrayExpression interface
+type ArrayExpression[E Expression] interface {
```
go-jet, line 4: Maybe just `Array[E Expression]`, so it follows naming convention with `Range`. Also `Expression` would repeat twice in the name.

internal/jet/array_expression.go
```diff
@@ -0,0 +1,93 @@
...
+	LT_EQ(rhs ArrayExpression[E]) BoolExpression
+	GT_EQ(rhs ArrayExpression[E]) BoolExpression
+
+	CONTAINS(rhs ArrayExpression[E]) BoolExpression
+	IS_CONTAINED_BY(rhs ArrayExpression[E]) BoolExpression
+	OVERLAP(rhs ArrayExpression[E]) BoolExpression
+	CONCAT(rhs ArrayExpression[E]) ArrayExpression[E]
+	CONCAT_ELEMENT(E) ArrayExpression[E]
+
+	AT(expression IntegerExpression) Expression
```
go-jet, line 20: `AT` should return element type `E`. Take a look at `rangeTypeCaster` how to extract element type.

internal/jet/expression.go
```diff
@@ -316,6 +316,32 @@ func (s *complexExpression) serialize(statement StatementType, out *SQLBuilder,
 	}
 }
 
+type arraySubscriptExpression struct {
```
go-jet, line 319: This is the old way to construct custom expression. The new and easier way is using `CustomExpression` and `Token`. In this case: `CustomExpression(array, Token("["), index, Token("]"))`.

internal/jet/literal_expression.go
```diff
@@ -160,6 +161,66 @@ func Decimal(value string) FloatExpression {
...
+
+	return &l
+}
+
+type unsafeArrayLiteral[E Expression] struct {
+	arrayInterfaceImpl[E]
+	literalExpressionImpl
+}
+
+func UnsafeArray[E LiteralExpression](values []interface{}) ArrayExpression[E] {
```
go-jet, line 216: Why is it needed?

internal/jet/literal_expression.go
```diff
@@ -160,6 +161,66 @@ func Decimal(value string) FloatExpression {
...
+
+	return &l
+}
+
+type integerArrayLiteral struct {
+	arrayInterfaceImpl[IntegerExpression]
+	literalExpressionImpl
+}
+
+func Int64Array(values []int64) ArrayExpression[IntegerExpression] {
```
go-jet, line 184: Can be done without new literal types. Some of these files need some cleaning.

```golang
func Int64Array(values []int64) ArrayExpression[IntegerExpression] {
    return ArrayExp[IntegerExpression](literal(pq.Int64Array(values))
}
```

internal/jet/sql_builder.go
```diff
@@ -253,6 +255,19 @@ func argToString(value interface{}) string {
 	}
 }
 
+func stringArrayQuote(val []string) string {
```
go-jet, line 258: Maybe we can use `pq.StringArray{}.Value()` instead, and cast first returned value to string.

internal/jet/string_expression.go
```diff
@@ -16,6 +16,9 @@ type StringExpression interface {
 	BETWEEN(min, max StringExpression) BoolExpression
 	NOT_BETWEEN(min, max StringExpression) BoolExpression
 
+	ANY_EQ(rhs ArrayExpression[StringExpression]) BoolExpression
+	ALL_EQ(rhs ArrayExpression[StringExpression]) BoolExpression
+
```
go-jet, line 21: If we add `ANY_EQ`, we would also need to add, for each of the types, NOT_EQ_ANY, LT_ANY, LT_EQ_ANY, etc... This would lead to number of methods explosion.

Simpler way would be to have `ANY` and `ALL` as generic functions:

```golang
func ANY[E Expression](array Array[E]) E {
    // check rangeTypeCaster
}
```

Then we can write for any type:

```golang
SomeStringColumn.EQ(ANY(Table.StringArrayColumn))
```

postgres/cast.go
```diff
@@ -45,6 +45,10 @@ type cast interface {
 	AS_INTERVAL() IntervalExpression
 }
 
+type castArray interface {
+	AS_STRING() jet.ArrayExpression[StringExpression]
```
go-jet, line 49: Why separate interface?

tests/postgres/scan_test.go
```diff
@@ -967,26 +969,25 @@ func TestScanIntoCustomBaseTypes(t *testing.T) {
 		ReplacementCost MyFloat64
 		Rating          *model.MpaaRating
 		LastUpdate      MyTime
-		SpecialFeatures *MyString
+		SpecialFeatures MyStringArray
```
go-jet, line 972: If `SpecialFeatures` is pq.StringArray you wouldn't need to exclude SpecialFeatures column bellow.

tests/postgres/alltypes_test.go
```diff
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"database/sql"
+	"github.com/lib/pq"
 	"testing"
 	"time"
```
go-jet, line 8: As for testing, maybe we can have a new array only table, similar to what's already been done for range tests.

go-jet: Note that test files are in separate repo: https://github.com/go-jet/jet-test-data

generator/template/model_template.go
arjen-ag5: You're right, but since we don't have a good model type yet, I haven't included this case yet

generator/template/sql_builder_template.go
arjen-ag5: I added a new `sqlArrayToColumnType` to convert a type to an array type. The function switches on the type name without the `[]` suffix because you would have to add infinite brackets for multidimensional arrays.

generator/template/sql_builder_template.go
arjen-ag5: The warning is handled in the caller

internal/jet/array_expression.go
arjen-ag5: 👍 Done

internal/jet/array_expression.go
arjen-ag5: Done

internal/jet/expression.go
arjen-ag5: Done

internal/jet/literal_expression.go
arjen-ag5: Idea was to support any type of array, but better not to include it

internal/jet/literal_expression.go
arjen-ag5: Cleanedup

internal/jet/sql_builder.go
arjen-ag5: Like this?
```
func stringArrayQuote(val []string) string {
	// We'll rely on the internals of pq2.StringArray here. We know it will never return an error, and the returned
	// value is a string
	dv, _ := pq2.StringArray(val).Value()
	return dv.(string)
}
```

internal/jet/string_expression.go
arjen-ag5: Yeah, that makes sense. I'll change that

generator/template/sql_builder_template.go
go-jet: When I think about it again it makes sense to separate arrays from `sqlToColumnType`. Since every postgres type can be element of array(`[]bool`, `[]point`, `[]timestamp`, etc...), sql builder array type can be constructed with just  `sqlToColumnType`:

```golang
columnType = sqlToColumnType(strings.TrimSuffix(typeName, "[]")) + "Array"
```

generator/template/sql_builder_template.go
go-jet: Aha, I see. Maybe we an error would be more in go style.

internal/jet/sql_builder.go
go-jet: Yeah, exactly.

arjen-ag5: Done

internal/jet/string_expression.go
arjen-ag5: Done

postgres/cast.go
arjen-ag5: Was left over cruft from experimentation, removed it

generator/template/sql_builder_template.go
arjen-ag5: How about?

```
// sqlToColumnType maps the type of a SQL column type to a go jet sql builder column. The second return value returns 
// whether the given type is supported.
func sqlToColumnType(typeName string) (string, bool) {
```

generator/template/sql_builder_template.go
arjen-ag5: This can lead to incorrectly generated code, because this PR does not support DateArray for example. What's your opinion on this?

tests/postgres/scan_test.go
arjen-ag5: With the type alias like so:
```
type MyStringArray pq.StringArray
```

The MyStringArray type does not implement Scanner. The QRM will then see it as a slice because that's what `pq.StringArray` is and try to apply it's row mapping into the slice. Therefore it will always be null.

generator/template/sql_builder_template.go
go-jet: It can be supported with one line `type DateArray Array[DateExpression]`. For model type we can fallback to string if there is no `pq` date array type.

generator/template/sql_builder_template.go
go-jet: Yeah, that is fine as well.

tests/postgres/scan_test.go
go-jet: No, I'm meant without defining a type, either by realiasing or not using new type at all:

```golang
type MyStringArray = pq.StringArray // note '=', MyStringArray now retains Scan method
```
or
```golang
	type film struct {
		FilmID          MyUint16 `sql:"primary_key"`
		Title           MyString
		Description     *MyString
		ReleaseYear     *MyInt16
		LanguageID      MyUint8
		RentalDuration  MyUint8
		RentalRate      MyFloat32
		Length          *MyUint32
		ReplacementCost MyFloat64
		Rating          *model.MpaaRating
		LastUpdate      MyTime
		SpecialFeatures *MyString
		SpecialFeatures pq.StringArray
	}
```

internal/jet/func_expression.go
```diff
@@ -651,6 +651,76 @@ func LEAST(value Expression, values ...Expression) Expression {
...
+
+// ANY should be used in combination with a boolean operator. The result of ANY is "true" if any true result is obtained
+func ANY[E Expression](arr Array[E]) E {
+	return arrayElementTypeCaster(arr, Func("ANY", arr))
+}
+
+// ALL should be used in combination with a boolean operator. TThe result of ALL is “true” if all comparisons yield true
+func ALL[E Expression](arr Array[E]) E {
+	return arrayElementTypeCaster(arr, Func("ALL", arr))
+}
```
go-jet, line 664: Sorry, I've made a mistake here. `ANY`/`ALL` should return BoolExpression.

internal/jet/func_expression.go
```diff
@@ -651,6 +651,76 @@ func LEAST(value Expression, values ...Expression) Expression {
...
+
+func ARRAY_APPEND[E Expression](arr Array[E], el E) Array[E] {
+	return arrayTypeCaster[E](arr, Func("array_append", arr, el))
```
go-jet, line 667: upercase "array_append".

postgres/literal.go
```diff
@@ -11,6 +11,11 @@ func Bool(value bool) BoolExpression {
 	return CAST(jet.Bool(value)).AS_BOOL()
 }
 
+// BoolArray creates new bool array literal expression
+func BoolArray(elements []bool) BoolArrayExpression {
+	return jet.BoolArray(elements)
```
go-jet, line 16: We'll probably need an explicit cast here as well, since cockroachdb tends to have problem with untyped parameters.

internal/jet/func_expression.go
go-jet: Ignore it.

sarthakjdev:
hey, would this be merged any time sooner? Have been using go-jet for a production project (https://wapikit.com) and just noticed, it does not supports arrays yet.

tylermmorton:
Hi I'm also interested in this functionality and I'm happy to push this work along if needed!

Do we know what needs to be done before this can be merged?

go-jet:
> Hi I'm also interested in this functionality and I'm happy to push this work along if needed!
> 
> Do we know what needs to be done before this can be merged?

Yeah, go ahead. To start, you can merge with the master branch and fix any failing tests. Then, go through the unresolved threads to identify what’s left to be done.

arjen-ag5:
Hello, I was out of it for a while, but I'll try to pick up work on this PR again

internal/jet/func_expression.go
arjen-ag5: I moved the ARRAY_APPEND, and other utilities to the postgres package because they are not compatible with MySQL or Sqlite

tests/postgres/alltypes_test.go
arjen-ag5: Done

internal/jet/func_expression.go
arjen-ag5: Done, and I since the ARRAY_ functions are postgres specific, I moved them to the postgres package

internal/jet/string_expression_test.go
```diff
@@ -76,6 +76,14 @@ func TestStringNOT_REGEXP_LIKE(t *testing.T) {
 	assertClauseSerialize(t, table3StrCol.NOT_REGEXP_LIKE(String("JOHN"), true), "(table3.col2 NOT REGEXP $1)", "JOHN")
 }
 
+func TestStringANY_EQ(t *testing.T) {
+	assertClauseSerialize(t, table2ColStr.EQ(ANY[StringExpression](table1ColStringArray)), "(table2.col_str = ANY(table1.col_array_string))")
```
go-jet, line 80: Can `ANY` work here without providing explicit type? 
```ANY(table1ColStringArray)```

File: .gitmodules [hunk 19]
```diff
@@ -1,3 +1,3 @@
 [submodule "tests/testdata"]
 	path = tests/testdata
-	url = https://github.com/go-jet/jet-test-data
+	url = https://github.com/arjen-ag5/jet-test-data
```
go-jet, line 3: This change will need to be part of separate pull request for https://github.com/go-jet/jet-test-data repo.

tests/postgres/array_test.go
```diff
@@ -0,0 +1,276 @@
...
+	. "github.com/go-jet/jet/v2/tests/.gentestdata/jetdb/test_sample/table"
+	"github.com/google/go-cmp/cmp"
+	"github.com/lib/pq"
+	"github.com/stretchr/testify/require"
+	"math/big"
+	"testing"
+)
+
+func TestArrayTableSelect(t *testing.T) {
+	skipForCockroachDB(t)
```
go-jet, line 17: Why skip for cockroachdb? cockroachdb supports arrays.

tests/postgres/array_test.go
```diff
@@ -0,0 +1,276 @@
...
+	query := SELECT(
+		SampleArrays.AllColumns,
+		SampleArrays.TextArray.EQ(SampleArrays.TextArray).AS("sample.text_eq"),
+		SampleArrays.BoolArray.EQ(boolArray).AS("sample.bool_eq"),
+		SampleArrays.TextArray.NOT_EQ(textArray).AS("sample.text_neq"),
+		SampleArrays.Int4Array.LT(int4Array).IS_TRUE().AS("sample.int4_lt"),
+		SampleArrays.Int8Array.LT_EQ(int8Array).IS_FALSE().AS("sample.int8_lteq"),
+		SampleArrays.TextArray.GT(textArray).AS("sample.text_gt"),
+		SampleArrays.Int4Array.GT_EQ(int4Array).AS("sample.bool_gteq"),
+		Int32(22).EQ(ANY[IntegerExpression](SampleArrays.Int4Array)).AS("sample.int4_eq_any"),
```
go-jet, line 33: Without explicit type?

go-jet:
Overall looks good.  I think we're close to finishing. There are only two things left:
1) Support for the rest of array types - FloatArray, TimeArray, TimezArray, DateArray, TimestampArray and TimestampzArray. I know there are no equivalent pq types, but we can go with supporting just sql builder. Model can remain string.
2) Support for cockroachdb array. There isn't much to be done here except include the same test data for cockroachdb as for postgres. I've noticed cockroachdb tests are failing, and it seems there are two issues at `all_types` table:
https://github.com/go-jet/jet-test-data/blob/1c501acb72bea389788404988ef0130b733f9cee/init/cockroach/test_sample.sql#L92-L96
* `integer` by default in cockroach is 64 bytes, so the fix is to force array to be array 32 byte integer `integer_array_ptr INT4[]`,
* `jsonb_array       text NOT NULL DEFAULT ''`, needs to become array `jsonb_array text[] NOT NULL DEFAULT array[]::text[]`,
Also I've made a one cockroach build fix, so you'll have to merge with master.

Bel1Vuk:
Hi, are there any updates or plans for merging this into master branch?

go-jet:
This feature will be included in the next release, expected sometime in autumn. If @arjen-ag5 is unavailable, I’ll take care of finalizing it.

noble-gase:
> This feature will be included in the next release, expected sometime in autumn. If @arjen-ag5 is unavailable, I’ll take care of finalizing it.


When will support ?

PR #387, opened 2024-09-18
Adding support for enum model AllValues

safaci2000:
Fixes #368

ChangeLog:
  - Updating test mysql version as it no longer exists.
  - Add a simple validation test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
           POSTGRES_DB: jetdb
           PGPORT: 50901
 
-      - image: circleci/mysql:8.0.27
+      - image: circleci/mysql:8.0
         command: [ --default-authentication-plugin=mysql_native_password ]
         environment:
           MYSQL_ROOT_PASSWORD: jet
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
           POSTGRES_DB: jetdb
           PGPORT: 50901
 
-      - image: circleci/mysql:8.0
+      - image: cimg/mysql:8.0
         command: [ --default-authentication-plugin=mysql_native_password ]
         environment:
           MYSQL_ROOT_PASSWORD: jet
@@ -25,7 +25,7 @@ jobs:
           MYSQL_PASSWORD: jet
           MYSQL_TCP_PORT: 50902
 
-      - image: circleci/mariadb:10.3
+      - image: cimg/mariadb:10.3
         command: [ '--default-authentication-plugin=mysql_native_password', '--port=50903' ]
         environment:
           MYSQL_ROOT_PASSWORD: jet
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
           POSTGRES_DB: jetdb
           PGPORT: 50901
 
-      - image: cimg/mysql:8.0
+      - image: circleci/mysql:8.0.27
         command: [ --default-authentication-plugin=mysql_native_password ]
         environment:
           MYSQL_ROOT_PASSWORD: jet
@@ -25,7 +25,7 @@ jobs:
           MYSQL_PASSWORD: jet
           MYSQL_TCP_PORT: 50902
 
-      - image: cimg/mariadb:10.3
+      - image: circleci/mariadb:10.3
         command: [ '--default-authentication-plugin=mysql_native_password', '--port=50903' ]
         environment:
           MYSQL_ROOT_PASSWORD: jet
--- a/generator/template/file_templates.go
+++ b/generator/template/file_templates.go
@@ -156,7 +156,7 @@ const (
 {{- end}}
 )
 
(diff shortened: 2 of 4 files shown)


safaci2000:
@go-jet if you have any suggestions on how to fix the circle_ci please let me know, I think this is a general overall issue with the code base rather than any changes I added, but I could be wrong.

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           POSTGRES_DB: jetdb
           PGPORT: 50901
 
-      - image: circleci/mysql:8.0.27
+      - image: cimg/mysql:8.0
```
go-jet, line 19: Why is it necessary?

generator/template/file_templates.go
```diff
@@ -156,6 +156,12 @@ const (
 {{- end}}
 )
 
+var {{$enumTemplate.TypeName}}_AllValues = []{{$enumTemplate.TypeName}} {
```
go-jet, line 159: Maybe without `_`.

go-jet:
> @go-jet if you have any suggestions on how to fix the circle_ci please let me know, I think this is a general overall issue with the code base rather than any changes I added, but I could be wrong.

The `circle_ci` is failing because some tests are failing - https://app.circleci.com/pipelines/github/go-jet/jet/825/workflows/587d624b-d435-4ef4-a168-9203c934932c/jobs/1059/tests

.circleci/config.yml
safaci2000: I'll revert the version change, I wasn't able to pull mysql:8.0.27, I assumed that was the case for circleci.  

Regarding the namespace, https://circleci.com/docs/circleci-images/

```
Legacy images with the prefix “circleci/” were [deprecated](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034)
 on December 31, 2021. For faster builds, upgrade your projects with [next-generation convenience images](https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic).
 ```

safaci2000: Actually, this may still be valid.  The CI and tests (docker-compose) can't point to the same version of MySQL since mysql:8.0.27 is no longer available while {cimg,circleci}/mysql:8.0.27 does exist.  Would you prefer both to be on 8.0 or leave the discrepancy?

generator/template/file_templates.go
safaci2000: Done

safaci2000:
@go-jet would you object to a patch that removes the need for external dependencies for tests? For example, I'm having a hard time getting the schema loaded into the MySQL server with the latest client. It would nice if that was simply using the go mysql driver to populate the data, same for the various other datastores.  I can add that to a different patch, but I think it'll make consistent testing easier.. wdyt?

.circleci/config.yml
go-jet: If you revert `.circleci/config.yml` and `docker-compose.yaml` the test should pass. `cimg/mysql:8.0` will download latest mysql version(v39) and tests are catching some differences in that version.

go-jet: I had no problem pulling `circleci/mysql:8.0.27`.

go-jet:
> @go-jet would you object to a patch that removes the need for external dependencies for tests? For example, I'm having a hard time getting the schema loaded into the MySQL server with the latest client. 

What issues do you have?

> It would nice if that was simply using the go mysql driver to populate the data, same for the various other datastores. I can add that to a different patch, but I think it'll make consistent testing easier.. wdyt?

Are you referring to this: https://github.com/go-jet/jet/blob/b835e25665decc06a6f3f4f2a3531aeff97c7f83/tests/init/init.go#L123

.circleci/config.yml
safaci2000: I reverted though keep in mind `docker pull mysql:8.0.27` from docker-compose will fail.

safaci2000:
> Are you referring to this:
> 
> https://github.com/go-jet/jet/blob/b835e25665decc06a6f3f4f2a3531aeff97c7f83/tests/init/init.go#L123

Yes, exactly.  The mysql is an external dependency and I was having issues with the server/client mismatch.  It would be nice if running the tests involved as few external dependencies as possible.  code is reproducible and in theory environment agnostic.  Same reason I like [testcontainers](https://testcontainers.com/).  When a developer runs go tests ./... it does exactly the same thing as the CICD, doesn't require any additional setup for the dev or the cicd.

go-jet:
> Yes, exactly. The mysql is an external dependency and I was having issues with the server/client mismatch. It would be nice if running the tests involved as few external dependencies as possible. code is reproducible and in theory environment agnostic. 

If I recall correctly, I've use mysql cli because initialization was much faster then when it is done from the code. But there is a chance I had something misconfigured in mysql.

> Same reason I like [testcontainers](https://testcontainers.com/). When a developer runs go tests ./... it does exactly the same thing as the CICD, doesn't require any additional setup for the dev or the cicd.
Maybe we should have a discussion topic about this? it's a unrelated to this PR.

Yeah, it makes sense. Lets start a separate discussion.

LGTM. :+1:

PR #389, opened 2024-09-20
Add Postgres DATE_TRUNC function

BranislavLazic:
Adds the `DATE_TRUNC` function for Postgres.

### DIFF
--- a/postgres/functions.go
+++ b/postgres/functions.go
@@ -332,6 +332,15 @@ var LOCALTIMESTAMP = jet.LOCALTIMESTAMP
 // NOW returns current date and time
 var NOW = jet.NOW
 
+// DATE_TRUNC returns the truncated date and time using optional time zone
+func DATE_TRUNC(field unit, source Expression, timezone ...string) TimestampExpression {
+	if len(timezone) > 0 {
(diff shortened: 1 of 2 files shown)


postgres/functions.go
```diff
@@ -332,6 +332,15 @@ var LOCALTIMESTAMP = jet.LOCALTIMESTAMP
 // NOW returns current date and time
 var NOW = jet.NOW
 
+// DATE_TRUNC returns the truncated date and time using optional time zone
+func DATE_TRUNC(field unit, source Expression, timezone ...string) TimestampExpression {
```
BranislavLazic, line 336: I'm not sure if we should return `TimestampzExpression` instead.

go-jet: Hmm, yeah it makes sense. `DATE_TRUNC` accepts any date/time type and can return either timestamp, timestamp with time zone, or interval. I guess timestamp is the most common use case, and the user can switch return type using `TimestampzExp` and `IntervalExp`. Could you just add note in comment for `TimestampzExp` and `IntervalExp`.

go-jet:
Thanks. Looks good. :+1:

Issue #391, opened 2024-09-24
Generate postgres comments


VolkerLieber:
It would be nice to also generate postgres comments.

This should be done by the referenced PR.
I'm not a postres internals expert, so please double check!

Greetings

PR #392, opened 2024-09-24
Include postgres comments in output

VolkerLieber:
#391

Please double check as I'm not an expert in postgres internals!


go-jet:
Nice. Didn't know postgres has comment support like mysql. 
Could you also include tests like the existing one for mysql: https://github.com/go-jet/jet/blob/6a0798eb06193b8833c7914fc09c257334520c52/tests/mysql/generator_test.go#L215
and
https://github.com/go-jet/jet/blob/6a0798eb06193b8833c7914fc09c257334520c52/tests/mysql/generator_test.go#L235

Note that test data files are in separate [repo](https://github.com/go-jet/jet-test-data).

VolkerLieber:
WIP: Tests still missing. Will add them asap

safaci2000:
There's also support for individual table column comments.  You could in theory add a // comment to each generated field if you like.

There's probably a much better way of doing this, but I wanted something that'll give you a single query for this.

here's the reference: https://www.postgresonline.com/journal/archives/215-Querying-table,-view,-column-and-function-descriptions.html



```sql
with table_comments as (
    SELECT n.nspname AS sname, c.relname As tname,
           CASE WHEN c.relkind = 'v' THEN 'view' ELSE 'table' END As type,
           d.description as table_comment
    FROM pg_class As c
             INNER JOIN pg_attribute As a ON c.oid = a.attrelid
             LEFT JOIN pg_namespace n ON n.oid = c.relnamespace
             LEFT JOIN pg_tablespace t ON t.oid = c.reltablespace
             LEFT JOIN pg_description As d ON (d.objoid = c.oid AND d.objsubid = 0) --
    WHERE c.relkind IN('r', 'v') -- AND d.description > ''
    ORDER BY n.nspname, c.relname
), column_comments as (
    SELECT n.nspname as sname, c.relname as tname , a.attname As column_name,  d.description as column_description
    FROM pg_class As c
             INNER JOIN pg_attribute As a ON c.oid = a.attrelid
             LEFT JOIN pg_namespace n ON n.oid = c.relnamespace
             LEFT JOIN pg_tablespace t ON t.oid = c.reltablespace
             LEFT JOIN pg_description As d ON (d.objoid = c.oid AND d.objsubid = a.attnum)
    WHERE  c.relkind IN('r', 'v') and d.description <> '' -- AND  n.nspname = 'SCHEMA' AND c.relname = 'TABLENAME'
    ORDER BY d.description, n.nspname, c.relname, a.attname
) select distinct t.sname, t.tname, t.table_comment, c.column_name, c.column_description from table_comments as t full outer join column_comments as c on t.tname = c.tname and t.sname = t.sname
where (t.table_comment <> '' or c.column_description <> '');
```

Should return the comments for both tables and columns.

VolkerLieber:
Thanks for the info :)
There seems to be a postgres function which I'm using in the latest PR.
I'll look into the url you provided tomorrow

VolkerLieber:
Waiting for https://github.com/go-jet/jet-test-data/pull/6

generator/metadata/enum_meta_data.go
```diff
@@ -1,7 +1,20 @@
...
+
+// GoLangComment returns enum comment without ascii control characters
+func (e Enum) GoLangComment() string {
+	if e.Comment == "" {
+		return ""
+	}
+
+	// remove ascii control characters from string
+	return regexp.MustCompile(`[[:cntrl:]]+`).ReplaceAllString(e.Comment, "")
 }
```
go-jet, line 20: This function repeats a lot. Maybe we can covert it to util function `GoLangComment(comment string) -> "// comment"` inside template package, and pass it to template engine using FuncMap, so we can write inside template:
``` 
{{GoLangComment .Comment}} 
type {{structImplName}} struct {
```

generator/postgres/query_set.go
```diff
@@ -14,7 +14,7 @@ type postgresQuerySet struct{}
 
 func (p postgresQuerySet) GetTablesMetaData(db *sql.DB, schemaName string, tableType metadata.TableType) ([]metadata.Table, error) {
 	query := `
-SELECT table_name as "table.name" 
+SELECT table_name as "table.name", obj_description((table_schema||'.'||quote_ident(table_name))::regclass) as "table.comment"
```
go-jet, line 17: Should we quote `table_schema` as well?

generator/postgres/query_set.go
```diff
@@ -57,6 +57,7 @@ func getColumnsMetaData(db *sql.DB, schemaName string, tableName string) ([]meta
 	query := `
 select  
     attr.attname as "column.Name",
+	dsc.description as "column.Comment",
```
go-jet, line 60: Since we are already using `obj_description` to get table column, maybe we should use `col_description` to get column description.

go-jet: Also note the ident, tabs vs spaces.

go-jet:
Thanks, looks good. :+1:

PR #397, opened 2024-09-29
Add Postgres GENERATE_SERIES function

BranislavLazic:
Adds GENERATE_SERIES function https://www.postgresql.org/docs/current/functions-srf.html
I could also add the GENERATE_SUBSCRIPTS function but it has to wait for array expressions https://github.com/go-jet/jet/pull/380.

### DIFF
--- a/postgres/functions.go
+++ b/postgres/functions.go
@@ -342,6 +342,15 @@ func DATE_TRUNC(field unit, source Expression, timezone ...string) TimestampExpr
 	return jet.NewTimestampFunc("DATE_TRUNC", jet.FixedLiteral(unitToString(field)), source)
 }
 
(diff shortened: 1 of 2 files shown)


go-jet:
Note that generate_series can be used as table source and expression:

```sql
SELECT * 
FROM generate_series(2,4); -- table source
```

```sql
SELECT generate_series(2,4); -- expression
```

With this PR the expression use case would be supported, but not the table source. Now to support table source it is a bit more work. I'm wondering if is it worth supporting the table source at all. At least to my knowledge everything that can be done with generate_series as a table source can be also done as an expression.

go-jet:
I guess we can go with just expression form, and revisit the other form if needed.

PR #400, opened 2024-10-02
adds columns list variable without default value columns

sirmackan:


sirmackan:
This allows for easy INSERTs without having to explicitly except common columns like "createdAt" which are often defined with a DEFAULT NOW() etc.

generator/template/file_templates.go
```diff
@@ -36,8 +36,9 @@ type {{structImplName}} struct {
 	{{$field.Name}} {{dialect.PackageName}}.Column{{$field.Type}} {{golangComment .Comment}}
 {{- end}}
 
-	AllColumns     {{dialect.PackageName}}.ColumnList
-	MutableColumns {{dialect.PackageName}}.ColumnList
+	AllColumns               {{dialect.PackageName}}.ColumnList
+	MutableColumns           {{dialect.PackageName}}.ColumnList
+	MutableNonDefaultColumns {{dialect.PackageName}}.ColumnList
```
go-jet, line 41: I'd rather go with `DefaultColumns` list. This way we can support two use cases:

```golang
Table.AllColumns.Except(Table.DefaultColumns)
// and
Table.MutableColumns.Except(Table.DefaultColumns)
```

sirmackan: That sounds good to me!

PR #404, opened 2024-10-05
Replacing several test util function with a generic version

safaci2000:
ChangeLog:
  - updated several test utils with a generic PtrOf
  - updated references using iotils (Deprecated) with os equivalent import.


safaci2000:
This is mostly fixing up the test tooling instead of using multiple implementation of BoolPtr, StringPtr etc, I removed all of them in favor of a PtroOf that uses generics.  Also, ioutils is deprecated so fixed up the imports.  All the functions have the same signature.

go-jet:
Good catch. To eliminate `testutils` from the name we can go with a new package `ptr` that will contain just one method `Of`, so we would write then `ptr.Of(false)`. Also this method isn't specific just for test, so we can move the package to https://github.com/go-jet/jet/tree/master/internal/utils.

safaci2000:
> Good catch. To eliminate `testutils` from the name we can go with a new package `ptr` that will contain just one method `Of`, so we would write then `ptr.Of(false)`. Also this method isn't specific just for test, so we can move the package to https://github.com/go-jet/jet/tree/master/internal/utils.

Done

go-jet:
LGTM. It seems some tests are now failing because of changes in jet-test-data.

go-jet:
> That's odd, that seems to go against how submodule are supposed to work. Do you mind if I just comment out the line:
> 
> ```
> checkout-testdata:
>    git submodule init
>    git submodule update
> # cd ./testdata && git fetch && git checkout master && git pull
> ```
> 
> if we're using submodule, it should have a commit ID that the testdata should honor. That line ignores all of it and always pulls master.

Yeah, you are right. Maybe move that line into `checkout-latest-testdat`.

safaci2000:
> Yeah, you are right. Maybe move that line into `checkout-latest-testdat`.

Done

PR #406, opened 2024-10-06
Improved support for intervals in postgres

### LINKED ISSUE #393: Better Support for Postgres Interval
1. When your schema uses postgres interval, Jet seems does not support doing an upsert (on conflict ) in postgres.

For example.

	insertStmt := funcapSchema.HtFleetStatus.
		INSERT(funcapSchema.HtFleetStatus.AllColumns).
		MODEL(req).
		ON_CONFLICT(funcapSchema.HtFleetStatus.RouterName).
		DO_UPDATE(SET(
			funcapSchema.HtFleetStatus.LastModified.SET(funcapSchema.HtFleetStatus.EXCLUDED.LastModified),
			funcapSchema.HtFleetStatus.UptimeValue.SET(funcapSchema.HtFleetStatus.EXCLUDED.UptimeValue),
		),
)

When UptimeValue is an interval, the SET command is not exposed, so it cannot be used as part of an upsert operation.

2. There should be no reason not to read in time.Interval as a time.Duration.  It's peculiar to have it read in as a string and then rely on the user to convert a very postgres specific format to a time stamp.

I'd like to have SET be available to pg interval and have the default data type be mapped to time.Duration.

**Work Around**

I'll post this in here for anyone who runs into this as well.

1. Code Generator

```go
	if column.DataType.Name == "interval" {
		defaultTableModelField.Type = template.NewType(new(time.Duration))
	}
```

2. Upsert

```go
	insertStmt := RawStatement(`
insert into funcap.ht_fleet_status (router_name, uptime_value, last_modified) values
    ($router_name, $interval::interval, now()) on conflict (router_name) do update
set uptime_value = excluded.uptime_value, last_modified = now() `,
		RawArgs{
			"$router_name": router,
			"$interval":    duration,
		},
	)
```

3. Query
```go
	stmt := SELECT(funcapSchema.HtFleetStatus.RouterName,
		//Raw required to allow use time.Duration in go model
		RawInt("(extract(epoch from ht_fleet_status.uptime_value)* 1000000000)::bigint").AS("ht_fleet_status.uptime_value"),
		funcapSchema.HtFleetStatus.LastModified,
	).FROM(funcapSchema.HtFleetStatus).WHERE(funcapSchema.HtFleetStatus.RouterName.EQ(String(routerName)))
```

safaci2000:
Fixes #393

### DIFF
--- a/tests/postgres/alltypes_test.go
+++ b/tests/postgres/alltypes_test.go
@@ -2,8 +2,10 @@ package postgres
 
 import (
 	"database/sql"
+	"github.com/go-jet/jet/v2/internal/jet"
 	"github.com/go-jet/jet/v2/internal/utils/ptr"
 	"github.com/stretchr/testify/assert"
+	"log/slog"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


safaci2000:
Blocked by testdata [PR](https://github.com/go-jet/jet-test-data/pull/7) and builds on top of https://github.com/go-jet/jet/pull/404.

tests/postgres/alltypes_test.go
```diff
@@ -128,6 +129,41 @@ WHERE all_types.uuid = 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11';
 	requireLogged(t, query)
 }
 
+func TestIntervalUpsert(t *testing.T) {
+	stmt := SELECT(Employee.AllColumns).FROM(Employee).
+		WHERE(Employee.EmployeeID.EQ(Int(1)))
+
+	var windy model.Employee
```
go-jet, line 136: SELECT seems redundant since we can use RETURNING.

tests/postgres/alltypes_test.go
```diff
@@ -128,6 +129,41 @@ WHERE all_types.uuid = 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11';
...
+	err := stmt.Query(db, &windy)
+	assert.Equal(t, windy.EmployeeID, int32(1))
+	assert.Equal(t, windy.FirstName, "Windy")
+	assert.Equal(t, windy.LastName, "Hays")
+	assert.Nil(t, err)
+	windy.PtoAccural = testutils.PtrOf("3h")
+	updateStmt := Employee.UPDATE(Employee.PtoAccural).SET(
+		Employee.PtoAccural.SET(CAST(String("3h")).AS_INTERVAL()),
```
go-jet, line 144: `CAST(String("3h")).AS_INTERVAL()` can be  `INTERVAL(3, HOURS)`

tests/postgres/alltypes_test.go
```diff
@@ -128,6 +129,41 @@ WHERE all_types.uuid = 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11';
...
+	).WHERE(Employee.EmployeeID.EQ(Int(1)))
+	_, err = updateStmt.Exec(db)
```
go-jet, line 146: updates and inserts should be executed inside transaction(and rollbacked) so we don't mess up test data. Check `testutils.ExecuteInTxAndRollback`

tests/postgres/alltypes_test.go
safaci2000: Done

tests/postgres/alltypes_test.go
```diff
@@ -931,6 +932,43 @@ func TestTimeExpression(t *testing.T) {
...
+			ON_CONFLICT(Employee.EmployeeID).
+			DO_UPDATE(SET(
+				Employee.PtoAccrual.SET(Employee.EXCLUDED.PtoAccrual),
+			)).RETURNING(Employee.AllColumns)
+		err = insertStmt.Query(db, &windy)
+		assert.Nil(t, err)
+		assert.Equal(t, *windy.PtoAccrual, "22:00:00")
+	})
+}
+
```
safaci2000, line 971: There might be an overall bug on how ExecuteInTrxAndRollback works, I had to make the final test use the original value so it won't affect other tests.  It doesn't seem to rollback the data.

go-jet: You need to pass `tx` to Query method for rollback to work:
```golang
insertStmt.Query(tx, &windy)
```

safaci2000: Doh.  Thanks.

tests/postgres/alltypes_test.go
```diff
@@ -931,6 +932,43 @@ func TestTimeExpression(t *testing.T) {
...
+		windy.PtoAccrual = ptr.Of("5h")
+		insertStmt := Employee.INSERT(Employee.AllColumns).
+			MODEL(&windy).
+			ON_CONFLICT(Employee.EmployeeID).
+			DO_UPDATE(SET(
+				Employee.PtoAccrual.SET(Employee.EXCLUDED.PtoAccrual),
+			)).RETURNING(Employee.AllColumns)
+		err = insertStmt.Query(tx, &windy)
+		assert.Nil(t, err)
+		assert.Equal(t, *windy.PtoAccrual, "05:00:00")
```
go-jet, line 968: We also need to make sure `SET' appears in the generated sql. Check `testutils.AssertStatementSql`.  
L937-L948 can be removed, since is unrelated to this test and there plenty of select scan tests. 
update and upsert are independent test case and can be grouped into separate sub tests `t.Run("update",...` and `t.Run("upsert"`.

safaci2000: done

tests/postgres/alltypes_test.go
```diff
@@ -931,6 +934,81 @@ func TestTimeExpression(t *testing.T) {
...
+			stmt := tc.statement(&windy)
+
+			testutils.AssertStatementSql(t, stmt, tc.expectedQuery)
+			err := stmt.Query(tx, &windy)
+			assert.Nil(t, err)
+			assert.Equal(t, *windy.PtoAccrual, tc.expectedInterval)
+
+		})
+	}
+}
```
go-jet, line 1010: I'm aware that table-driven tests tend to be endorsed by the golang community, but I'm mostly against them since the drawbacks far exceed the benefits. Especially for integration tests. Some of the drawbacks are:
- unreadable. I've seen tests with 1000 test cases, and I've ended up scrolling up and down, just to figure out the test logic. Table-driven tests usually take 3x,4x, or more lines of code than regular tests.
- complicates debugging. I can't just put a breakpoint at a specific test case, I have to run them all(I know some IDEs provide help in this regard).
- tests are rigid. All test cases can do only one thing. In this case, execute the statement and check the result. For instance, if I want to check a slightly different scenario that affects only let's say `insertQuery` I can't, I have to write a different query.

For simple unit tests that check just one function, the rigid point does not apply, but unreadability and complicated debugging are the reasons I tend to avoid them completely even for unit tests. 

In this case, two explicit sub tests would suffice:

```golang
func TestIntervalSetFunctionality(t *testing.T) {

t.Run("update", 
t.Run("upsert",

}
```

safaci2000: Oh, okay.  I misunderstood your request then.  Let me fix this up.

go-jet:
Thanks. Looks good. :+1:

PR #407, opened 2024-10-08
Fixing various security issues and updating null_types to handle over…

safaci2000:
Adding gosec and lint scanning to project


safaci2000:
@go-jet do you have any insights on this? I'm not very familiar with circleCI, but it seems like all the tests are passing, the tests tab is all clear, the report has no failures I can find.  Though it seems something is still breaking.

.circleci/config.yml
```diff
@@ -163,4 +163,4 @@ workflows:
   version: 2
   build_and_test:
     jobs:
-      - build_and_tests
\ No newline at end of file
+      - build_and_tests
```
go-jet, line 166: Whats the diff in this line? Is the new line missing?

qrm/internal/null_types.go
```diff
@@ -111,40 +112,58 @@ type NullUInt64 struct {
...
 	switch v := value.(type) {
 	case nil:
 		n.Valid = false
 		return nil
 	case int64:
-		n.UInt64, n.Valid = uint64(v), true
-		return nil
+		return handleSignedInts(validatePositive(v), func() {
+			n.UInt64, n.Valid = uint64(v), true // #nosec G115
+		})
```
go-jet, line 138: This method tends to be called a lot during QRM, so adding two additional method calls to check a sign might introduce some latency. To avoid it, a simple if check would be preferable.

qrm/internal/null_types_test.go
```diff
@@ -88,4 +101,29 @@ func TestNullUInt64(t *testing.T) {
...
+
+	//Validate negative use cases
+	err := nullUInt64.Scan(int64(-5))
+	assert.NotNil(t, err)
+	assert.Equal(t, err.Error(), bufferOverflowError.Error())
+
+	//Validate negative use cases
+	err = nullUInt64.Scan(-5)
+	assert.NotNil(t, err)
+	assert.Equal(t, err.Error(), bufferOverflowError.Error())
```
go-jet, line 113: You can use `assert.Error` instead.

qrm/internal/null_types.go
```diff
@@ -111,40 +112,58 @@ type NullUInt64 struct {
...
 }
 
+// validatePositive validates that the numeric value of any time is a positive entity, safely casts to unsigned type
+func validatePositive[T constraints.Integer](n T) bool {
+	return n > 0
+}
+
 // Scan implements the Scanner interface.
 func (n *NullUInt64) Scan(value interface{}) error {
+	var badState = fmt.Errorf("cannot cast a negative value to an unsigned value, buffer overflow error")
```
go-jet, line 122: Error can be moved outside the function, so it can be used inside tests for comparison.

.circleci/config.yml
safaci2000: That wasn't intentional, removed.  Was trying to see what was off with the tests failing.

qrm/internal/null_types.go
safaci2000: Okay, that's fair.  Though there is a legit risk of buffer overflow.  If you cast -5 you get something like 18446744073709551611 (Depending on which data type you end up with).  I can just have a lot of repetitive code if you prefer.  

I updated the PR to reflect that, let me know if that's what you had in mind?

go.mod
```diff
@@ -1,23 +1,21 @@
...
 require (
 	github.com/go-sql-driver/mysql v1.8.1
+	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
 	github.com/jackc/pgconn v1.14.3
-	github.com/lib/pq v1.10.9
-	github.com/mattn/go-sqlite3 v1.14.17
-)
-
-require (
```
go-jet, line 13: The idea between the two groups was that one is for the library dependencies and the second is for test dependencies.

internal/testutils/test_utils.go
```diff
@@ -103,21 +102,11 @@ func AssertJSON(t *testing.T, data interface{}, expectedJSON string) {
...
-// SaveJSONFile saves v as json at testRelativePath
-func SaveJSONFile(v interface{}, testRelativePath string) {
-	jsonText, _ := json.MarshalIndent(v, "", "\t")
-
-	filePath := getFullPath(testRelativePath)
-	err := os.WriteFile(filePath, jsonText, 0644)
-
-	throw.OnError(err)
-}
-
```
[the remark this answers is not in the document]
go-jet: It can stay, just in case if we need to update some json result files.

qrm/internal/null_types.go
go-jet: Yeah, looks good.

go-jet:
> @go-jet do you have any insights on this? I'm not very familiar with circleCI, but it seems like all the tests are passing, the tests tab is all clear, the report has no failures I can find. Though it seems something is still breaking.

Hmm, it is weird. Indeed, all tests pass. A wield guess would be that after all the tests have finished, `TestMain' panics at some point. But only `profile delay.Start().Stop()` is executed afterwards.

Maybe add some log lines and panic catcher in `TestMain`.
If that fails, then I guess start removing the changes one by one until we find out exactly which change is breaking the tests.

go.mod
safaci2000: Rolled back. Does go differentiate between the two?  Unless it's just for readability? My understanding that the dependencies included are based on the imports actually invoked in the dependency graph.  ie. 2 different main.go would have 2 different dependency trees.

safaci2000:
> > @go-jet do you have any insights on this? I'm not very familiar with circleCI, but it seems like all the tests are passing, the tests tab is all clear, the report has no failures I can find. Though it seems something is still breaking.
> 
> Hmm, it is weird. Indeed, all tests pass. A wield guess would be that after all the tests have finished, `TestMain' panics at some point. But only `profile delay.Start().Stop()` is executed afterwards.
> 
> Maybe add some log lines and panic catcher in `TestMain`. If that fails, then I guess start removing the changes one by one until we find out exactly which change is breaking the tests.

The issue is with go-junit-report.  I replaced with a very similar tool that I think is a bit easier to use.  It also, actually gives you a break down as the tests are running without having to navigate to the tests tab.

Let me know if you have any objections to using this instead, otherwise I'll try to figure out how to make it go-junit-report work.

Also, how would you feel about moving the circle CI tests into a github action ?

go.mod
go-jet: Just for readability.

tests/mysql/main_test.go
```diff
@@ -33,7 +30,6 @@ func sourceIsMariaDB() bool {
 }
 
 func TestMain(m *testing.M) {
-	rand.Seed(time.Now().Unix())
```
go-jet, line 36: Why?

go-jet:
> > > @go-jet do you have any insights on this? I'm not very familiar with circleCI, but it seems like all the tests are passing, the tests tab is all clear, the report has no failures I can find. Though it seems something is still breaking.
> > 
> > 
> > Hmm, it is weird. Indeed, all tests pass. A wield guess would be that after all the tests have finished, `TestMain' panics at some point. But only `profile delay.Start().Stop()` is executed afterwards.
> > Maybe add some log lines and panic catcher in `TestMain`. If that fails, then I guess start removing the changes one by one until we find out exactly which change is breaking the tests.
> 
> The issue is with go-junit-report. I replaced with a very similar tool that I think is a bit easier to use. It also, actually gives you a break down as the tests are running without having to navigate to the tests tab.
> 
> Let me know if you have any objections to using this instead, otherwise I'll try to figure out how to make it go-junit-report work.
> 
> Also, how would you feel about moving the circle CI tests into a github action ?

What was the issue with go-junit-report?

I'm fine with gotestsum, but I don't see codecoverage on [codecov](https://app.codecov.io/gh/go-jet/jet/commits/All%20branches) now

tests/mysql/main_test.go
safaci2000: It's deprecated and no longer needed, from go docs.

```
// Deprecated: As of Go 1.20 there is no reason to call Seed with
// a random value. Programs that call Seed with a known value to get
// a specific sequence of results should use New(NewSource(seed)) to
// obtain a local random generator.
```

safaci2000:
> I'm fine with gotestsum, but I don't see codecoverage on [codecov](https://app.codecov.io/gh/go-jet/jet/commits/All%20branches) now.
> 
> I'm fine with github actions as well, if they provide the same as circle.ci.

Okay, let me see if I can get the junit report to work, or fix gotestsum.  They should be compatible... let me see what I can do with that.  

Not 100% sure of the issue, but it required an update to V2 and it still didn't work for me.  I can't replicate it locally so need to use the CICD to test the behavior.

safaci2000:
> I'm fine with github actions as well, if they provide the same as circle.ci.

Okay, I'll do another MR without any other changes being added for that later on if I can manage something that's comparable.

safaci2000:
Okay, done.  Fixed the code coverage issue.

go-jet:
> Okay, done. Fixed the code coverage issue. I created a token for my own project and verified the upload. https://app.codecov.io/gh/safaci2000/jet/tree/feature%2Fscanners

Something ain't right. Code coverage should be above 90%. Also the last two steps `store_artifacts` and `store_test_results` did not upload anything.

tests/mysql/main_test.go
go-jet: Aha, I see.

safaci2000:
> > Okay, done. Fixed the code coverage issue. I created a token for my own project and verified the upload. https://app.codecov.io/gh/safaci2000/jet/tree/feature%2Fscanners
> 
> Something ain't right. Code coverage should be above 90%. Also the last two steps `store_artifacts` and `store_test_results` did not upload anything.

Noted, I'll look at it when i have a chance.

safaci2000:
I had to bump the go version used for testing a bit further than go.mod due to a change in behavior. https://github.com/golang/go/issues/65653. 

Code coverage should be resolved

go-jet:
> I had to bump the go version used for testing a bit further than go.mod due to a change in behavior. [golang/go#65653](https://github.com/golang/go/issues/65653).

Interesting, now it makes sense.

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   build_and_tests:
     docker:
       # specify the version
-      - image: cimg/go:1.21.6
+      - image: cimg/go:1.22.2
```
go-jet, line 11: Lets then use latest 22 version, just to avoid any other potential bugs.

.circleci/config.yml
```diff
@@ -128,17 +128,13 @@ jobs:
...
 
-      # this will run all tests and exclude test files from code coverage report
-      - run:  |
-          go test -v ./... \
-            -covermode=atomic \
-            -coverpkg=github.com/go-jet/jet/v2/postgres/...,github.com/go-jet/jet/v2/mysql/...,github.com/go-jet/jet/v2/sqlite/...,github.com/go-jet/jet/v2/qrm/...,github.com/go-jet/jet/v2/generator/...,github.com/go-jet/jet/v2/internal/... \
-            -coverprofile=cover.out 2>&1 | go-junit-report > $TEST_RESULTS/results.xml
+      - run:
+          name: Running tests
+          command:  gotestsum --junitfile report.xml --format testname   -- -coverprofile=cover.out  -covermode=atomic -coverpkg=github.com/go-jet/jet/v2/postgres/...,github.com/go-jet/jet/v2/mysql/...,github.com/go-jet/jet/v2/sqlite/...,github.com/go-jet/jet/v2/qrm/...,github.com/go-jet/jet/v2/generator/...,github.com/go-jet/jet/v2/internal/... ./...
```
go-jet, line 137: report.xml -> $TEST_RESULTS/results.xml

.github/workflows/code_scanner.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+env:
+  GO_VERSION: 1.22.0
```
go-jet, line 11: Lets avoid this buggy version.

.github/workflows/code_scanner.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  security_scanning:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: 1.22.0
```
go-jet, line 28: As well.

go.mod
```diff
@@ -1,19 +1,19 @@
 module github.com/go-jet/jet/v2
 
-go 1.18
+go 1.22.0
```
go-jet, line 3: Why is it necessary to increase the minimum go version?

.circleci/config.yml
safaci2000: Done.

.circleci/config.yml
safaci2000: Done

.github/workflows/code_scanner.yml
safaci2000: Done

go.mod
safaci2000: Let me try to bump that down, there is a pattern where setup-go can simply pull the version from the go.mod which is convenient, but not having to force everyone to upgrade is valid too.  Let me see if this works with 1.18.

safaci2000: Yup, this works.  I think it got updated when did a go install of gosec.  Reverted.

go-jet:
Thanks. Looks good. :+1:

PR #418, opened 2024-10-27
Fix generating schema fails due to column name conflicts

### LINKED ISSUE #408: Generating schema Fails
**Describe the bug**
The code generator should fail if datamodel contains reserved fields.  

There is a collision with the way the code is constructed, because postgres.Table exposes TalbeName() and the table contains a field named TableName the code does not compile.

**Environment (please complete the following information):**
 - OS: mac os X
 - Database: postgres
-  Database driver: pgx
 - Jet version: master

**Code snippet**

```sql
create schema audit;
drop table if exists audit.audit_actions;
create table audit.audit_actions
(
    schema_name text not null,
    table_name text not null,
    trx_id text default txid_current(),
    user_name text,
    app_user text,
    action TEXT NOT NULL,
    old_data jsonb,
    new_data jsonb,
    query text,
    creation_date timestamp with time zone DEFAULT now() NOT NULL
) with (fillfactor=100);
```

**Expected behavior**
Generated code to compile successfully or fail hard on the generation and provide context on the reason for the failure

Neuman968:
fixes #408 

- Added a function to rename a column if it conflicts with reserved keywords. 

- Updated sql_builder_template to check if column is in a conflict with reserved key words

- Added unit tests. 

Thank you for this awesome project! Let me know if there is anything I can add or change.

### DIFF
--- a/generator/template/sql_builder_template.go
+++ b/generator/template/sql_builder_template.go
@@ -5,6 +5,7 @@ import (
 	"github.com/go-jet/jet/v2/generator/metadata"
 	"github.com/go-jet/jet/v2/internal/utils/dbidentifier"
 	"path"
+	"slices"
 	"strings"
 	"unicode"
 )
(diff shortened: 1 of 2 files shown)


go-jet:
Thanks. Looks good. :+1:

PR #419, opened 2024-11-01
Add PostgreSQL-specific character type constructors: Text, Char, and VarChar

go-jet:


alexandrkara-outreach:
Why was `postgres.Uint64()` removed from postgres/literal.go? We depend on it.

go-jet:
> Why was `postgres.Uint64()` removed from postgres/literal.go? We depend on it.

The reasoning was that since postgres does not support unsigned types, passing very large values might introduce silent overflow bugs. But now when I double checked postgres does return an error in such cases. So I guess it's safe. It's reintroduced on master branch again.

alexandrkara-outreach:
Thank you!

Issue #413, opened 2024-10-21
Add go_package cmd param


DrBlury:
**I need to generate my go structs in a different package. The package name is repo.**
Currently the generator will always generate with package name model, enum or table.

Solution (preferred):
I'd like another parameter to be added that allows me to define the go package name for the generated files. (Or multiple parameters for each package)

`jet -dsn=postgres://$APP_DATABASE_USER:$APP_DATABASE_PASSWORD@localhost:$APP_DATABASE_PORT/$APP_DATABASE_NAME?sslmode=disable -schema=public -path=./.gen`

-gopkgmodel=entities, -gopkgtable=something, -gopkgenum=dbenums 

Something like this would help a lot!

https://github.com/go-jet/jet/issues/208 does not solve my issue. I think this should be easily accessible via CLI.

safaci2000:
I'm not sure I understand why that particular behavior matters, then again, I don't think there's anything preventing that feature.  

That being said....

@go-jet what do you think about simply exposing the 'templates' as files the user can override?  

Right now, file_templates.go contains all these static data.  You can use embed.FS and have a fallback to read from a predefined directory like jet_templates/table.tmpl.go 

It would allow users to add their own customizations without being forced to commit the change upstream and wait for a release?

go-jet:
> @go-jet what do you think about simply exposing the 'templates' as files the user can override?

Exposing template file will make them part of the public API, which entails a commitment to maintain and support them unchanged until the next major release. There are also numerous `FuncMap` functions, not just static data. Exposing these internal implementation details could limit future refactoring options, as changes would need to avoid breaking the public contract. If a developer needs to customize the template, they always have the option to fork the project and implement any necessary modifications.

jamius19:
I think this would be a good addition to the command line parameters.

@go-jet I can take a look into this if you're fine with adding some optional parameters to the cli

jamius19:
Alright, I'll start working on it.
I'll submit a PR when it's done!

jamius19:
Do these CLI flags work?

1. `-model-pkg` for the **model** package name
2. `-table-pkg` for the **table** package name
3. `-view-pkg` for the **view** package name
4. `-enum-pkg` for the **enum** package name

jamius19:
I've added a PR #424, please review at your convenience.

jamius19:
The flags are `rel-model-path`, `rel-table-path`, `rel-view-path`, `rel-enum-path`.
This issue can now be closed @go-jet

[...]

PR #424, opened 2024-11-21
Added CLI Flags for Package name customization for Model, Table, View and Enum

jamius19:
### Goal
Jet doesn't currently support custom package names for the generated files for Models, Tables, Views, and Enums. It always uses the fixed `model`, `table`, `view`, and `enum` names for these packages. This PR will allow users to pass **OPTIONAL** CLI arguments to Jet for custom package names.

### Features Added
Users can customize the package names through these flags,

1. `-model-pkg` for the model package name
2. `-table-pkg` for the table package name
3. `-view-pkg` for the view package name
4. `-enum-pkg` for the enum package name

If any or all of these flags are absent, the default package names (e.g., `model` for Model) will be used making this a **NON Breaking** change.

I've also added a corresponding integration test for these optional CLI arguments in Postgres. This should be sufficient as all databases use the `main.go:genTemplate()` function that was changed to accommodate this feature.

For additional context, please see Issue #413


cmd/jet/main.go
```diff
@@ -71,6 +75,10 @@ func init() {
 	flag.StringVar(&ignoreEnums, "ignore-enums", "", `Comma-separated list of enums to ignore`)
 
 	flag.StringVar(&destDir, "path", "", "Destination dir for files generated.")
+	flag.StringVar(&modelPkg, "model-pkg", "", "Package name for the model files generated")
```
go-jet, line 78: We should be able to pass a relative path not just package name. 
Maybe usage should say something like:  `Model files package relative path to destination directory`.
Also default value is missing.

cmd/jet/main.go
```diff
@@ -243,10 +252,16 @@ func genTemplate(dialect jet.Dialect, ignoreTables []string, ignoreViews []strin
 		return strslice.Contains(ignoreEnums, strings.ToLower(enum.Name))
 	}
 
+	model := template.DefaultModel()
+
+	if modelPkg != "" {
+		model = model.UsePath("/" + modelPkg)
+	}
```
go-jet, line 259: If you add default flag values above, this can be simplified:
```golang
template.DefaultModel().UsePath(modelPkg) // I believe there is no need for "/"
```

tests/postgres/generator_test.go
```diff
@@ -338,6 +398,33 @@ var MpaaRating = &struct {
...
 
+var mpaaRatingEnumFileWithPkgName = `
+//
+// Code generated by go-jet DO NOT EDIT.
+//
+// WARNING: Changes to this file may cause incorrect behavior
+// and will be lost if the code is regenerated
+//
+
+package newenum
```
go-jet, line 409: There might be no need for the new test variables. We should be able to reuse existing one by just replacing `package enum` with `package newenum`.

cmd/jet/main.go
jamius19: Initially, I was only thinking about allowing the customization of the package name.
But yeah, this would make it more flexible.

Will do.

cmd/jet/main.go
jamius19: Noted.

tests/postgres/generator_test.go
jamius19: Okay, I'll replace the line `package enum` with the new package from existing variables then.

cmd/jet/main.go
jamius19: There is indeed no need for the `/` as you noted as all `Path` are being used with `path.Join()`

cmd/jet/main.go
jamius19: Descriptions were incorrect as well, fixed.

jamius19: Changing the description of `destDir` to replace `Destination dir` with `Destination directory` for consistency with the new flags.

jamius19: Adding a period at the end of the usage description of `-ignore-tables`, `-ignore-views`, and `-ignore-enums` flags to make them consistent with all the other flags.

jamius19:
Rebased my branch with `master`

jamius19:
@go-jet Please don't merge this PR yet, relative directory names starting with `./` or `.\` and nested directory names don't work on Windows. Most probably related to Issue #426

I'll investigate and update here.

jamius19:
Issue #426 is related to how PowerShell handles arguments.
TL;DR Arguments should be passed as quoted strings in PowerShell.

However, if a quoted nested directory is passed as the value of `-model-pkg`, it breaks on Windows.
I'm checking up on it now.

jamius19:
Fixed path compatibility issues on Windows.

jamius19:
Since we're passing a path for the model package, wouldn't it be better if the flag was `model-pkg-dir`?

Same goes for the other three as well.

go-jet:
> Since we're passing a path for the model package, wouldn't it be better if the flag was `model-pkg-dir`?
> 
> Same goes for the other three as well.

Shouldn't be `path` then (`model-pkg-path`)? Since directory is technically just a container.
Maybe we can also drop `pkg`(it kind of implies), and add `rel`. - `rel-model-path`, `rel-table-path`, `rel-enum-path`, `rel-view-path`

jamius19:
I agree with both points that it should be `path` and that the `pkg` is redundant.
I also think the addition of `rel` in these new flags will help indicate that they are relative to something (the `path` flag in this case).

I'm changing the flags like this `-rel-model-path` then.

cmd/jet/main.go
```diff
@@ -66,17 +71,28 @@ func init() {
...
-func main() {
 	flag.Usage = usage
 	flag.Parse()
 
+	// Convert the OS-specific path separator to slashes for cross-platform compatibility.
+	destDir = filepath.ToSlash(destDir)
+	modelPkg = filepath.ToSlash(modelPkg)
+	tablePkg = filepath.ToSlash(tablePkg)
+	viewPkg = filepath.ToSlash(viewPkg)
+	enumPkg = filepath.ToSlash(enumPkg)
```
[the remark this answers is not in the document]
jamius19: This wouldn't be needed if the following method in the `Model` struct used `filepath.Base()` instead of `path.Base()`,

```go
// model_template.go:24
// PackageName returns package name of model types
func (m Model) PackageName() string {
	return path.Base(m.Path)
}
```

The `path` package doesn't work correctly with backslashes ([source](https://pkg.go.dev/path))

> Package path implements utility routines for manipulating slash-separated paths.
>
>The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the [path/filepath](https://pkg.go.dev/path/filepath) package.

And almost all the path-related operations (those I've seen at least) in Jet are done using the `path` package.

\
To illustrate the problem,
```go
fmt.Println(path.Base( ".\\entity"))             // .\entity
fmt.Println(path.Base(".\\test\\entity"))       // .\test\entity
fmt.Println(path.Base("./entity"))             // entity
```

Whereas all of these paths correctly return `entity` as the base when using the `filepath.Base()` function.
\
As a result of this, when Jet is invoked with the model path flag set to `.\entity` on Windows, the generated model's package statement becomes `package .\entity` because it's getting the package name from the `PackageName()` method in the `process.go:processTableModels()` function,
```go
text, err := generateTemplate(
	// ...
	template.FuncMap{
		"package": func() string {
			return modelTemplate.PackageName()
		},
		// ...
})
```

And the `format.Source(text)` call in `filesys.go:FormatAndSaveGoFile()` fails because of the invalid package name.

\
Admittedly my change is more of a workaround. A better approach would be to migrate all the file path related operations from using the `path` package to the `filepath` package.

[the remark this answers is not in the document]
jamius19: I think changing all the usage  of the `path` package to the `filepath` package is the right approach.
I've looked into all the usage of `path` and it should be relatively simple and non breaking.

Shall I go along with that?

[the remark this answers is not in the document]
jamius19: As it turns out, there were actually not that many `path` package usage. So I fixed them.
Works on Windows without the replacing slash workaround now.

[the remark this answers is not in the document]
go-jet: > I think changing all the usage of the path package to the filepath package is the right approach.
I've looked into all the usage of path and it should be relatively simple and non breaking.

Yeah, it makes sense.

cmd/jet/main.go
```diff
@@ -170,6 +178,7 @@ func usage() {
 		"source", "dsn", "host", "port", "user", "password", "dbname", "schema", "params", "sslmode",
 		"path",
 		"ignore-tables", "ignore-views", "ignore-enums",
+		"model-pkg", "table-pkg", "view-pkg", "enum-pkg",
```
go-jet, line 181: You'll need to update this names as well.

jamius19: Yeah, my bad.
Pushed the changes already.

go-jet: Great. LGTM. :+1:

Issue #423, opened 2024-11-20
Local test suite is broken because of mismatched MySQL/MariaDB version with CircleCI


jamius19:
**Describe the bug**
MySQL and MariaDB versions are not the same between `.circleci/config.yml` and `tests/docker-compose.yaml`, and as such, the local test suite fails. 

The output from `go test -v ./...` run from the repository root is attached here,
[jet-test-verbose.log](https://github.com/user-attachments/files/17835062/jet-test-verbose.log)

And the output of `make jet-gen-all` if needed,
[jet-make-gen-all.log](https://github.com/user-attachments/files/17835064/jet-make-gen-all.log)

\
**Environment (please complete the following information):**
 - OS: `linux (Ubuntu 22.04.5 LTS, 6.8.0-49-generic # 49~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Nov  6 17:42:15 UTC 2 x86_64 x86_64 x86_64 GNU/Linux)`
 - Database: `mysql/maridb`
-  Database driver: `N/A`
 - Jet branch/version: `master / v2.11.1`

**Code snippet**
N/A

**Expected behavior**
The test suite would PASS on the master branch as it does on CircleCI

**Possible Solution**
Replacing the MySQL and MariaDB versions in the `tests/docker-compose.yaml` with the ones from `.circleci/config.yml` fixes the test suite. Only changing both versions to the ones from CircleCI's config seems to fix it, although they should be in sync regardless.

I already have a local branch fixing this. Should I open a PR?

go-jet:
Yep, go ahead.

PR #425, opened 2024-11-21
Fixed local test suite by syncing MySQL and MariaDB versions from CircleCI config

jamius19:
### Goal
MySQL and MariaDB versions are not the same between `.circleci/config.yml` and `tests/docker-compose.yaml`, and as such, the local test suite fails on the `master` branch whereas it passes in CircleCI.

I've synced versions of MySQL and MariaDB databases so that local tests work correctly.
Please see Issue #423 for more context.


go-jet:
Thanks. Looks good.

Issue #433, opened 2024-12-17
add authorization to always quoted


Form1ca:
**Describe the bug**
wrong sql generated if column name **authorization**
PostgreSQL
<nil> pq: syntax error at or near "authorization"

**Code snippet**
```
INSERT INTO table (authorization)
VALUES (NULL);
``` 

**Expected behavior**
```
INSERT INTO table ("authorization")
VALUES (NULL);
```

iyiola-dev:
hi, i think this can be avoided by not using postgres keywords as table names.

go-jet:
Fixed in `Release v2.13.0`.

PR #434, opened 2024-12-23
Add postgres keyword authorization

formica2:
fix bug
https://github.com/go-jet/jet/issues/433


go-jet:
Why closed? Looks good.

Form1ca:
> Why closed? Looks good.

this will fix my case, but I can’t guarantee that it won’t break something

https://www.postgresql.org/docs/current/sql-keywords-appendix.html
AUTHORIZATION  reserved (**can be function or type**)


https://www.postgresql.org/docs/current/sql-set-session-authorization.html
```
ok
RESET SESSION AUTHORIZATION

error
RESET SESSION `AUTHORIZATION`
```

I considered the options:
- put all column names in quotes
- add a flag to the column settings: put in quotes

formica2:
add all reserved keyword
from https://www.postgresql.org/docs/current/sql-keywords-appendix.html

go-jet:
Thanks, looks good. :+1:

PR #436, opened 2024-12-26
feat: add generator skip option for table columns

### LINKED ISSUE #435: Generator: Add option to Skip columns/fields

I'm adapting an existing database structure using the jet generator. It works fine but it generates the tables with all columns/fields.
I would like to be able to skip columns/fields from tables as well in the generator.
There doesn't seem to be a way to skip columns/fields using the `TableSQLBuilder.UseColumn` function like is possible with schemas/tables.

Adding a `Skip` option to the `TableSQLBuilderColumn` structure (`UseColumn(columnsFunc func(column metadata.Column) TableSQLBuilderColumn)`) so columns/fields can be programmatically skipped.

Is this something that is already possible? If not, would this be something that could be implemented?

galexrt:
This is a quick implementation of having a skip option for table columns/fields in the generator for the feature request I created #435.
This might not be optimal but the easiest way I found to implement this feature.

Resolves #435


go-jet:
Hi @galexrt, 
Yeah, it makes sense to have such a feature.
Not sure if it is intentional, with this change you're skipping only SQL builder columns, not model fields.
Could you also add a test here: https://github.com/go-jet/jet/blob/f9ed359785fcff71fbf1e48dfa3b8f5862d98302/tests/postgres/generator_template_test.go#L451

galexrt:
I have added a test and due to it being easier to handle skipped fields for the `column-list`, I had added a template function `columnList` to render the list via `strings.Join`.

go-jet:
Thanks, looks good. :+1:

PR #443, opened 2025-01-17
generator/*: allow passing DB directly

### LINKED ISSUE #441: generator/sqlite: add function to generate using an existing *sql.DB
I want to apply my migrations to an in-memory sqlite db and then generate the go-jet structures using that. However, right now it's not possible as GenerateDSN takes a connection string instead of an existing DB connection. My workaround is to write the DB to a temporary file and use that. I am aware that I can make the DB global in-memory to the process and then connect to it. However, this ruins isolation from other DBs in the process.

I also use alternative sqlite implementations like "modernc.org/sqlite" which are registered as "sqlite" instead of "sqlite3" which the GenerateDSN expects so I have to reregister the driver I want as "sqlite3" before calling the generator.

Add the following function that behaves the same as GenerateDSN except it applies directly to the supplied *sql.DB.

```go
func GenerateDB(db *sql.DB, destDir string, templates ...template.Template) error
```

TotallyGamerJet:
Closes #441

### DIFF
--- a/generator/postgres/postgres_generator.go
+++ b/generator/postgres/postgres_generator.go
@@ -56,6 +56,14 @@ func GenerateDSN(dsn, schema, destDir string, templates ...template.Template) er
 	defer db.Close()
 
 	fmt.Println("Retrieving schema information...")
+	return GenerateDB(db, schema, cfg.Database, destDir, templates...)
+}
+
+// GenerateDB generates jet files using the provided *sql.DB
+func GenerateDB(db *sql.DB, dbName, schema, destDir string, templates ...template.Template) error {
+	if dbName == "" {
+		return fmt.Errorf("database name is required")
+	}
 	generatorTemplate := template.Default(postgres.Dialect)
 	if len(templates) > 0 {
 		generatorTemplate = templates[0]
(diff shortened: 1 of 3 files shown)


generator/postgres/postgres_generator.go
```diff
@@ -56,6 +56,14 @@ func GenerateDSN(dsn, schema, destDir string, templates ...template.Template) er
 	defer db.Close()
 
 	fmt.Println("Retrieving schema information...")
+	return GenerateDB(db, schema, cfg.Database, destDir, templates...)
```
go-jet, line 59: The schema name and database name are swapped here, which is causing the tests to fail.

generator/postgres/postgres_generator.go
```diff
@@ -66,7 +74,7 @@ func GenerateDSN(dsn, schema, destDir string, templates ...template.Template) er
 		return fmt.Errorf("failed to get '%s' schema metadata: %w", schema, err)
 	}
 
-	dirPath := filepath.Join(destDir, cfg.Database)
+	dirPath := filepath.Join(destDir, dbName)
```
go-jet, line 77: Hmm, I don’t see much value in passing two separate strings to functions only to have them joined within the function. It seems better to join them into `destDir` in the caller instead, so `dbName` can be removed from this method.

go-jet:
Thanks. Looks good. :+1:

PR #451, opened 2025-02-06
DefaultColumns() helper on Table

sirmackan:
See https://github.com/go-jet/jet/pull/400

### DIFF
--- a/generator/metadata/table_meta_data.go
+++ b/generator/metadata/table_meta_data.go
@@ -21,3 +21,16 @@ func (t Table) MutableColumns() []Column {
 
 	return ret
 }
+
+// DefaultColumns returns list of columns with default values set for table
+func (t Table) DefaultColumns() []Column {
+	var ret []Column
+
+	for _, column := range t.Columns {
+		if column.HasDefault {
+			ret = append(ret, column)
+		}
+	}
+
+	return ret
+}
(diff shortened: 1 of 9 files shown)


go-jet:
`TestGenerator` for mariadb is failing. Note that tests in `/tests/mysql` are run twice, once for mysql and once for mariadb.

sirmackan:
Not sure what to do though

go-jet:
> Not sure what to do though

I see, MySQL and MariaDB view information schema are different. In my opinion, both behaviors are incorrect since views don’t have the concept of a default column value like table columns do. I think we should follow Postgre's behavior and avoid marking view columns as default.  

At this line:
https://github.com/go-jet/jet/blob/00b8155f74dbeb51f5f5c7351f002c632a22f401/generator/mysql/query_set.go#L21
adding `AND t.table_type != 'VIEW'` should be enough.

sirmackan:
Did what you suggested @go-jet  👍

go-jet:
Thanks. Looks good. :+1:

JonasRothmann:
I think this creates some issues relating to `GENERATED ALWAYS AS IDENTITY` in postgresql as it is not covered in defaultColumns

go-jet:
> I think this creates some issues relating to GENERATED BY DEFAULT AS IDENTITY in postgresql as it is not covered in defaultColumns. Do we want to adapt the query_set to consider those as default too?

Yeah, it makes sense to include identity columns as well.

Issue #483, opened 2025-04-10
CLI: Add option to skip model and/or SQL builder generation


joshkaplinsky:
**Is your feature request related to a problem? Please describe.**

No current issues, functionality already exists in the generator, this request is to expand support for existing generator options through the CLI.

**Describe the solution you'd like**

We use a mix of SQLC and Jet in our projects and both provide the ability to generate models. Depending on our use case, we'd like the ability to skip generating models (or SQLBuilder) entirely via the CLI w/o requiring us to build our own generator. The functionality already exists to skip specific tables/views/enums, but it would be great to be able to skip models or SQL builder entirely.

go-jet:
Makes sense. 👍

PR #484, opened 2025-04-10
feat: add support for skipping model and sql builder in CLI

joshkaplinsky:
Closes https://github.com/go-jet/jet/issues/483.

This PR expands the CLI to support skipping model and/or SQLBuilder generators entirely.

### DIFF
--- a/generator/template/model_template_test.go
+++ b/generator/template/model_template_test.go
@@ -1,11 +1,9 @@
 package template
 
 import (
-	"testing"
-
-	"github.com/stretchr/testify/require"
-
 	"github.com/go-jet/jet/v2/generator/metadata"
+	"github.com/stretchr/testify/require"
+	"testing"
 )
 
 func Test_TableModelField(t *testing.T) {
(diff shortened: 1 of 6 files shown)


generator/template/model_template_test.go
```diff
@@ -43,3 +45,29 @@ func Test_TableModelField(t *testing.T) {
 		Tags: nil,
 	})
 }
+
+func Test_Model_ShouldSkip(t *testing.T) {
```
go-jet, line 49: This test doesn't add much value, as it just sets a field and then reads the same value that was assigned.
Take a look at [generator_test.go](https://github.com/go-jet/jet/blob/master/tests/postgres/generator_test.go).

generator/template/sql_builder_template_test.go
```diff
@@ -35,3 +37,29 @@ func TestColumnRenameReserved(t *testing.T) {
 		})
 	}
 }
+
+func Test_SQLBuilder_ShouldSkip(t *testing.T) {
```
go-jet, line 41: The same.

generator/template/model_template_test.go
joshkaplinsky: Will remove, agreed that they add very little value. Added these since the gocov check was failing after the patch.

go-jet: > Added these since the codecov check was failing after the patch.

I see. In that case new test in [generator_template_test.go](https://github.com/go-jet/jet/blob/master/tests/postgres/generator_template_test.go) would be a better option, since `generator_test.go` is testing jet generator executable and thus codecov is not included in the result.

joshkaplinsky: The functionality I'm introducing is already covered by existing tests in `generator_template_test.go`, since I'm just externalizing the `Skip` boolean to the CLI. For example:

https://github.com/go-jet/jet/blob/6c1224017dcdcd54036da3d80902d5aa996eaa19/tests/postgres/generator_template_test.go#L58-L73

Happy to add a test to [generator_test.go](https://github.com/go-jet/jet/blob/6c1224017dcdcd54036da3d80902d5aa996eaa19/tests/postgres/generator_test.go), if you'd like, despite it not being included in codecov.

go-jet: Updating any existing tests in `generator_template_test.go` to include `ShouldSkip` method call should be enough for codecov to pass.

go-jet:
Thanks. Looks good. :+1:

PR #495, opened 2025-04-28
feat: add preliminary json support for qrm_pgx.

DevX86:


DevX86:
@go-jet Something like this?

DevX86:
Self reminder on the `stmt.QueryPGX()` to avoid the
```go
query, args := stmt.Sql()
_, err := qrm.QueryPGX(ctx, r.db, query, args, row)
if err != nil {
	return err
}
	
return nil
```

We want to make a `statement.go` type pgx implementation but also wanted to avoid `PGX` spam on the `stmt.` intellisense.
Two ideas were put forth, `stmt.PGX.Query()` and `pgx.Query()` personally, leading towards the latter, less dots and it gives things a more modular development experience.

go-jet:
> @go-jet Something like this?

Yeah, looks good.

qrm/qrm_pgx.go
```diff
@@ -8,9 +8,63 @@ import (
 	"reflect"
 )
 
-// QueryablePGX interface for pgx Query method
+// QueryablePGX interface for pgx QueryContext method
 type QueryablePGX interface {
-	Query(ctx context.Context, sql string, args ...any) (pgx.Rows, error)
+	QueryContext(ctx context.Context, query string, args ...any) (pgx.Rows, error)
```
go-jet, line 13: Does pgx have `QueryContext` method?

go-jet: This interface is for `pgx` types. Since there is no `QueryContext` method in `pgx`, the build is failing:

```sh
=== Errors
tests/postgres/pgx_test.go:34:49: cannot use pgxDB (variable of type *pgx.Conn) as qrm.QueryablePGX value in argument to qrm.QueryPGX: *pgx.Conn does not implement qrm.QueryablePGX (missing method QueryContext)
```

Issue #216, opened 2023-02-20
Whitelist flags for generate command


K1DV5:
**Is your feature request related to a problem? Please describe.**
I wanted to generate models for a few tables in a big database. I understand that there is the `-ignore-tables` flag, but it is too much work to list the 45 tables to generate models for 5 tables in a 50 table database. And if a new table is added to the database and I don't want to include it in the models, I would have to remember to include it as the 46th table to ignore.

**Describe the solution you'd like**
I think it'd be great if there was something like `-allow-tables` flag which would take precedence over `-ignore-tables` when given. And same applies about enums and views.

go-jet:
Yeah, it makes sense. At the moment this behavior can be also achieved using generator customization([model example](https://github.com/go-jet/jet/blob/f2e4b8551c48b97d0cd2d3deff47dc1b2aa2f04e/tests/postgres/generator_template_test.go#L173), [sql builder example](https://github.com/go-jet/jet/blob/f2e4b8551c48b97d0cd2d3deff47dc1b2aa2f04e/tests/postgres/generator_template_test.go#L201)).

PR #506, opened 2025-06-18
Added whitelist flags for generate command

Sanjaiy:
Added whitelist flags for generate command. Related to [issue](https://github.com/go-jet/jet/issues/216).
**Flags**
  -allow-tables
	Comma-separated list of tables to allow. Takes precedence over --ignore-tables flag.
  -allow-views
	Comma-separated list of views to allow. Takes precedence over --ignore-views flag.
  -allow-enums
	Comma-separated list of enums to allow. Takes precedence over --ignore-enums flag.

### DIFF
--- a/cmd/jet/main.go
+++ b/cmd/jet/main.go
@@ -51,14 +51,14 @@ var (
 	viewPkg  string
 	enumPkg  string
 
-	allowTables string
-	allowViews  string
-	allowEnums  string
+	tables string
+	views  string
+	enums  string
 )
 
 type templateFilter struct {
 	names []string
-	allow bool
+	ignore bool
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/jet/main.go
```diff
@@ -84,6 +93,10 @@ func init() {
 	flag.StringVar(&tablePkg, "rel-table-path", "table", "Relative path for the Table files package from the destination directory.")
 	flag.StringVar(&viewPkg, "rel-view-path", "view", "Relative path for the View files package from the destination directory.")
 	flag.StringVar(&enumPkg, "rel-enum-path", "enum", "Relative path for the Enum files package from the destination directory.")
+
+	flag.StringVar(&allowTables, "allow-tables", "", `Comma-separated list of tables to allow. Takes precedence over --ignore-tables flag.`)
+	flag.StringVar(&allowViews, "allow-views", "", `Comma-separated list of views to allow. Takes precedence over --ignore-views flag.`)
+	flag.StringVar(&allowEnums, "allow-enums", "", `Comma-separated list of enums to allow. Takes precedence over --ignore-enums flag.`)
```
go-jet, line 99: Not sure about `allow` word in the name. Maybe just `tables`, `views` and `enums`.

go-jet: Also, if the user uses both options (`-tables` and `-ignore-tables`), it would be best to return an error rather than silently choose one option.

cmd/jet/main.go
```diff
@@ -50,8 +50,17 @@ var (
...
 	enumPkg  string
+
+	allowTables string
+	allowViews  string
+	allowEnums  string
 )
 
+type templateFilter struct {
+	names []string
+	allow bool
```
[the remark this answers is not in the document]
Sanjaiy: @go-jet Makes sense. Refactored the logic to use `ignore bool`.

Issue #510, opened 2025-07-18
When using table.FromSchema() with postgres, schemas starting with a number are not quoted


desol:
**Describe the bug**
If I have a schema named 1test, then the serializer for table names does not identify that the schema should be quoted for postgres.

**Environment (please complete the following information):**
 - OS: linux
 - Database: postgres
-  Database driver: pqx
 - Jet version 2.13.0

**Code snippet**
	customerTable := Customers.FromSchema("1test")

**Expected behavior**
	generates: select * from "1test".users instead of select * from 1test.users

go-jet:
Code of [interest](https://github.com/go-jet/jet/blob/8667bca8fc898299b72102bd1564c8c4fbd5af3c/internal/jet/sql_builder.go#L304) and [test](https://github.com/go-jet/jet/blob/8667bca8fc898299b72102bd1564c8c4fbd5af3c/internal/jet/sql_builder_test.go#L53).

PR #511, opened 2025-07-20
fix: quote identifiers starting with numbers

Andrei-hub11:
**Problem:** Schema names starting with digits (e.g., `1test`) were not being properly quoted, generating invalid SQL like `FROM 1test.users` instead of `FROM "1test".users`.

**Solution:** Enhanced `shouldQuoteIdentifier()` function to detect identifiers starting with digits using `unicode.IsNumber()` and added empty string validation to prevent panics.

Closes: #510

### DIFF
--- a/internal/jet/sql_builder.go
+++ b/internal/jet/sql_builder.go
@@ -4,15 +4,16 @@ import (
 	"bytes"
 	"database/sql/driver"
 	"fmt"
-	"github.com/go-jet/jet/v2/internal/3rdparty/pq"
-	"github.com/go-jet/jet/v2/internal/utils/is"
-	"github.com/google/uuid"
 	"reflect"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
 	"unicode"
+
+	"github.com/go-jet/jet/v2/internal/3rdparty/pq"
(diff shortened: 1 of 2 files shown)


internal/jet/sql_builder.go
```diff
@@ -302,12 +303,21 @@ func integerTypesToString(value interface{}) string {
...
+		return true
+	}
+
 	_, err := strconv.ParseInt(identifier, 10, 64)
 
 	if err == nil { // if it is a number we should quote it
 		return true
 	}
 
+	firstChar := rune(identifier[0])
```
go-jet, line 316: I don't think this will get the first rune. This will only cast the first byte(uint8) to rune(uint32).  
To get the first rune you'd have to first cast to rune slice: `[]rune(identifier)[0]`.

But, I think the cleaner approach would be to use existing for loop bellow and check if index is 0(there is existing number check).

Andrei-hub11: Thank you for letting me know about this. I pushed a new commit trying to follow the suggested approach.

go-jet:
Thanks, looks good. :+1:

Issue #486, opened 2025-04-11
Add option to Jet CLI to generate models with JSON struct tags


IcyFoxe:
I tried to figure out how to generate code with JSON struct tags, but all solutions lead to this link at https://github.com/go-jet/jet/wiki/Generator#generator-customization, which as far as I understand means you need to generate from code instead from prebuilt CLI, and you need to implement it yourself, which is not an easy task for someone who's just starting out with this package.

It would be nice if there was a simple option in Jet CLI that would add these JSON struct tags to the models automatically. For example `-json-tags=true`.

go-jet:
Agree, makes sense to add a new cli option. Something like: `-model-json-tag` with possible values `none`, `camel-case`, `pascal-case` or `snake-case`.

mispp:
@go-jet how would you propagate this flag until DefaultTableModelField?

carsonkrueger:
@go-jet I've opened a new PR fixing this issue #516

go-jet:
Support added in the Release v.2.14.0.

PR #516, opened 2025-08-04
add model json tag to args

carsonkrueger:
- Add flag to cmd to allow table/view models to generate json tags using snake, camel, or pascal case.
- Add tests for both table and models and for each casing

### DIFF
--- a/cmd/jet/main.go
+++ b/cmd/jet/main.go
@@ -366,7 +366,7 @@ func createModelTags(columnMetaData metadata.Column) []string {
 	var tags []string
 	switch modelJsonTag {
 	case "snake-case":
-		tags = append(tags, fmt.Sprintf(`json:"%s"`, columnMetaData.Name))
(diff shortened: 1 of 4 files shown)


carsonkrueger:
fixes #486

cmd/jet/main.go
```diff
@@ -344,3 +361,16 @@ func shouldSkipEnum(enum metadata.Enum, filter templateFilter) bool {
 
 	return !strslice.Contains(filter.names, strings.ToLower(enum.Name))
 }
+
+func createModelTags(columnMetaData metadata.Column) []string {
+	var tags []string
+	switch modelJsonTag {
+	case "snake-case":
+		tags = append(tags, fmt.Sprintf(`json:"%s"`, columnMetaData.Name))
```
go-jet, line 369: If a user chooses  `snake-case` then database names are probably in `camel-case`. 
`SnakeToCamel` function is originally copied from this library - https://github.com/serenize/snaker to avoid including the whole library. We can do the same for `CamelToSnake` function.

go-jet:
Thanks. Looks good. :+1:

PR #520, opened 2025-08-23
Add ability to skip cleaning the output directory

tylermmorton:
I've run into a snag recently while using Jet's postgres generator. It completely cleans the output directory when it goes to process code generation on a database schema. 

The way that my project is set up, I want to generate multiple schemas into the same directory while also having each schema be in its own subpackage:
```
   /storage
     /schema
       /public
       /computed
       /telemetry
```

Unfortunately I can't do this when I call `ProcessDB` multiple times, because it ends up clearing the `/schema` folder of the files it just generated. 

I tried my best to make the proposed changes backwards compatible but I'm open to feedback. I'll maintain a fork of jet for now but I'd love to see this feature merged.

Thanks

### DIFF
--- a/generator/postgres/postgres_generator.go
+++ b/generator/postgres/postgres_generator.go
@@ -79,6 +79,54 @@ func GenerateDB(db *sql.DB, schema, destDir string, templates ...template.Templa
 	return nil
 }
 
+type GeneratorOption func(opts *generatorOptions)
+
+type generatorOptions struct {
+	DestDir   string
+	SkipClean bool
+}
+
+func WithSkipClean() GeneratorOption {
+	return func(opts *generatorOptions) {
+		opts.SkipClean = true
+	}
+}
+
+func WithDestDir(destDir string) GeneratorOption {
+	return func(opts *generatorOptions) {
+		opts.DestDir = destDir
+	}
+}
+
+func GenerateWithOptions(db *sql.DB, schema string, templates []template.Template, options ...GeneratorOption) error {
+	generatorTemplate := template.Default(postgres.Dialect)
+	if len(templates) > 0 {
+		generatorTemplate = templates[0]
+	}
+
+	schemaMetadata, err := metadata.GetSchema(db, &postgresQuerySet{}, schema)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


tylermmorton:
> > because it ends up clearing the /schema folder of the files it just generated.
> 
> This shouldn't be the case. If you regenerate `public` schema `computed` and `telemetry` should remain intact.
> 
> How are you invoking the generator?

I'm invoking the generator using Go in a mage script. I only want to generate the SQL builders into this package.
```go
	for _, schemaName := range store.SchemaNames {
		err = postgres.Generate(db, schemaName,
			[]template.Template{
				template.Default(postgres2.Dialect).
					UseSchema(func(schemaMetaData metadata.Schema) template.Schema {
						var skipModel = template.DefaultModel()
						skipModel.Skip = true
						return template.DefaultSchema(schemaMetaData).
							UsePath(filepath.Join("schema")).
							UseModel(skipModel).
							UseSQLBuilder(template.DefaultSQLBuilder().
								UseEnum(func(enum metadata.Enum) template.EnumSQLBuilder {
									skipEnum := template.DefaultEnumSQLBuilder(enum)
									skipEnum.Skip = true
									return skipEnum
								}).
								UseView(func(table metadata.Table) template.ViewSQLBuilder {
									skipView := template.DefaultViewSQLBuilder(table)
									skipView.Skip = true
									return skipView
								}).
								UseTable(func(table metadata.Table) template.TableSQLBuilder {
									skipTable := template.DefaultTableSQLBuilder(table)
									skipTable.Skip = true
									if slices.Contains(skipTables[:], table.Name) {
										return skipTable
									}

									return template.DefaultTableSQLBuilder(table).
										UsePath(schemaName).
										UseFileName(fmt.Sprintf("%s%s.go", strman.SnakeLower(table.Name), genSuffix)).
										UseColumn(func(column metadata.Column) template.TableSQLBuilderColumn {
											return template.DefaultTableSQLBuilderColumn(column)
										})
								}),
							)
					}),
			}
		)
		if err != nil {
			return fmt.Errorf("failed to generate Jet code for schema %s: %w", schemaName, err)
		}
	}
```

I had played around with calling `UsePath("/")` on the sql builder template but it hits an error.

PR #523, opened 2025-10-08
WIP: Add support for pgx v5 interface

go-jet:
How to use:
```golang
err = pgxV5.Query(ctx, query, pgxDB, &result)
```
where `pgxDB` can be `*pgx.Tx`, `*pgxpool.Pool` or `*pgx.Conn`.

Default jet generator will work in most cases, except for some postgres specific types like `numeric`, `bit`, `point`, etc...
For those types you'll need to customize generator to use `pgtype.Numeric`, `pgtype.Bits`, `pgtype.Point`, etc...
The full list of such types can be found in this [test](https://github.com/go-jet/jet/blob/48b7a20a31ae89d3dba071955bf129466d89a7db/tests/postgres/pgx_test.go#L168).

The reason why it haven't been merged to master and released is:
- to make it work for uuid types I had to add a hack into hot qrm path. Which I'm not happy about, and this will affect needlessly none pgx users. 
- to get a complete performance improvements of pgx driver probably a bigger refactor of qrm is  required.
- as described above default generator misses a lot of type.

All the tests are passing and you can give it a try.

Even if pgx branch doesn't get into next release I'll keep it in sync with master.


DevX86:
Thank you so much for this @go-jet I'm reviewing this and the adapter thread now.

go-jet:
> Works great so far! We may have covered this in the past but I'm trying to refresh my memory. Any reason for pgxV5.Query() instead of stmt.QueryPGX() off of the original statement builder I rather have the new one import style in your latest PR than the previous multiline qrm style but figured I'd bring it up. I believe it had something todo with a statement.go needing to be implemented but wanted to get your blessing before trying things.

The reason is because `pgx` is already on version 5 and what guaranties us there want be tomorrow a version 6, 7, 8 etc... I felt this pollutes statements API.

> Doing SELECT_JSON_ARR() with pgxV5.Query() works but for some reason a time.Time on a created_at is giving me the nil timestamp.

Does it work with standard `database/sql` driver? If not, then it is probably something related to a query or a destination.
Could you share your destination?

PR #526, opened 2025-10-22
Stabilize import order

shu-bc:
Import order of generated model files changes from time to time. It causes a large number of file changes in my current working project. Stable import order might be solution for this.

Running `gofmt` after generation may be an alternative, but I don't think anyone could be in favor of unstable import order.

### DIFF
--- a/generator/template/model_template.go
+++ b/generator/template/model_template.go
@@ -2,11 +2,12 @@ package template
 
 import (
 	"fmt"
-	"github.com/lib/pq"
 	"path/filepath"
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/lib/pq"
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgtype"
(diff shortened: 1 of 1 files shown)


generator/template/model_template.go
```diff
@@ -112,21 +113,17 @@ func (t TableModel) UseField(structFieldFunc func(columnMetaData metadata.Column
...
+	seen := map[string]bool{}
+	var ret []string
 	for _, columnMetaData := range tableMetaData.Columns {
 		field := modelType.Field(columnMetaData)
 		for _, importPath := range append([]string{field.Type.ImportPath}, field.Type.AdditionalImportPaths...) {
-			if importPath != "" {
-				importPaths[importPath] = true
+			if importPath != "" && !seen[importPath] {
+				ret = append(ret, importPath)
 			}
```
go-jet: Not sure, how this change would stabilize import order??

PR #534, opened 2025-12-05
fix: correct ColumnAssigment typo to ColumnAssignment

joshua-temple:
## Summary
- Rename `ColumnAssigment` to `ColumnAssignment` (correct spelling)
- Add deprecated type alias `ColumnAssigment` for backwards compatibility
- Rename internal types and variables to use correct spelling
- Rename files: `column_assigment.go` -> `column_assignment.go`

## Backwards Compatibility
The old `ColumnAssigment` type remains available as a deprecated type alias pointing to `ColumnAssignment`. Existing code will continue to work without changes.


internal/jet/column_assignment.go
```diff
@@ -0,0 +1,23 @@
+package jet
```
go-jet: Could you use `git mv` for file renames. Simple rename loses git history.

PR #536, opened 2025-12-09
JSON select support for sqlite

oscar345:
Adds support for selecting json with sqlite. To create tests, I have copied the tests from mysql and added them to sqlite test folder. I made some tweaks to the tests, for example using a different test database for sqlite and making small changes to the sql syntax for the asserts. This PR also changes the qrm package for dealing with returned strings instead of bytes for json values in sqlite.

There are two tests I do not get working, and those are:

`TestSelectJson_GroupBy`: the returned json from sqlite has different values for items in the amount key. Some sums or averages are slightly different because of rounding errrors, but it does mean the test wont pass. For example: 

Original from test
```json
{
		"CustomerID": 2,
		"StoreID": 1,
		"FirstName": "PATRICIA",
		"LastName": "JOHNSON",
		"Email": "PATRICIA.JOHNSON@sakilacustomer.org",
		"AddressID": 6,
		"Active": "1",
		"CreateDate": "2006-02-14T22:04:36Z",
		"LastUpdate": "2019-04-11T18:11:49Z",
		"Amount": {
			"Sum": *128.73000000000002*,
			"Avg": *4.767777777777779*,
			"Max": 10.99,
			"Min": 0.99,
			"Count": 27
		}
```

JSON from sqlite json statement:
```json
{
  "CustomerID": 2,
  "StoreID": 1,
  "FirstName": "PATRICIA",
  "LastName": "JOHNSON",
  "Email": "PATRICIA.JOHNSON@sakilacustomer.org",
  "AddressID": 6,
  "Active": "1",
  "CreateDate": "2006-02-14T22:04:36Z",
  "LastUpdate": "2019-04-11T18:11:49Z",
  "Amount": {
    "Sum": *128.73*,
    "Avg": *4.76777777777778*,
    "Max": 10.99,
    "Min": 0.99,
    "Count": 27
  }
}
```

The other test that fails needs more work, but i have not been able to solve it. The `TestSelectJsonObj_NestedObj` returns json but the nested json statements are not returned as json by sqlite but as strings containing json (so it is not a nested object but a string). Because of that the json unmarshal function does not work. To make sqlite return a nested object instead of a string, a JSON() statement should be added that wraps the subquery. So the query that fails is:

```sql
SELECT JSON_OBJECT(
          'actorID', actor.actor_id,
          'firstName', actor.first_name,
          'lastName', actor.last_name,
          'lastUpdate', strftime('%Y-%m-%dT%H:%M:%fZ', actor.last_update),
          'LongestFilm', (
               SELECT JSON_OBJECT(
                         'filmID', film.film_id,
                         'title', film.title,
                         'description', film.description,
                         'releaseYear', film.release_year,
                         'languageID', film.language_id,
                         'originalLanguageID', film.original_language_id,
                         'rentalDuration', film.rental_duration,
                         'rentalRate', film.rental_rate,
                         'length', film.length,
                         'replacementCost', film.replacement_cost,
                         'rating', film.rating,
                         'specialFeatures', film.special_features,
                         'lastUpdate', strftime('%Y-%m-%dT%H:%M:%fZ', film.last_update)
                    ) AS "json"
               FROM film_actor
                    INNER JOIN film ON (film.film_id = film_actor.film_id)
               WHERE actor.actor_id = film_actor.actor_id
               ORDER BY film.length DESC
               LIMIT ?
               OFFSET ?
          )
     ) AS "json"
FROM actor
WHERE actor.actor_id = ?;
```

But to make it work, the query should look like this:

```sql
SELECT JSON_OBJECT(
          'actorID', actor.actor_id,
          'firstName', actor.first_name,
          'lastName', actor.last_name,
          'lastUpdate', strftime('%Y-%m-%dT%H:%M:%fZ', actor.last_update),
          'LongestFilm', JSON((  -- <- notice the json function here
               SELECT JSON_OBJECT(
                         'filmID', film.film_id,
                         'title', film.title,
                         'description', film.description,
                         'releaseYear', film.release_year,
                         'languageID', film.language_id,
                         'originalLanguageID', film.original_language_id,
                         'rentalDuration', film.rental_duration,
                         'rentalRate', film.rental_rate,
                         'length', film.length,
                         'replacementCost', film.replacement_cost,
                         'rating', film.rating,
                         'specialFeatures', film.special_features,
                         'lastUpdate', strftime('%Y-%m-%dT%H:%M:%fZ', film.last_update)
                    ) AS "json"
               FROM film_actor
                    INNER JOIN film ON (film.film_id = film_actor.film_id)
               WHERE actor.actor_id = film_actor.actor_id
               ORDER BY film.length DESC
               LIMIT ?
               OFFSET ?
          ))
     ) AS "json"
FROM actor
WHERE actor.actor_id = ?;
```

### DIFF
--- a/tests/sqlite/select_json_test.go
+++ b/tests/sqlite/select_json_test.go
@@ -3,7 +3,6 @@ package sqlite
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -565,8 +564,6 @@ func TestSelectJson_QueryWithoutUnMarshaling(t *testing.T) {
 		).AS("raw_json"),
 	)
 
-	fmt.Println(stmt.DebugSql())
-
 	testutils.AssertDebugStatementSql(t, stmt, strings.ReplaceAll(`
 SELECT (
           SELECT JSON_GROUP_ARRAY(JSON_OBJECT(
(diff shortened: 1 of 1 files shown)


sqlite/dialect.go
```diff
@@ -29,6 +30,24 @@ func newDialect() jet.Dialect {
...
+			// CustomExpression used bellow (instead DATE_FORMAT function) so that only expr is parametrized
+			case TimestampExpression:
+				return CustomExpression(Token("strftime('%Y-%m-%dT%H:%M:%fZ', "), e, Token(")"))
+			case TimeExpression:
+				return CustomExpression(Token("strftime('%Y-%m-%dT%H:%M:%fZ', "), e, Token(")"))
+			case DateExpression:
+				return CustomExpression(Token("strftime('%Y-%m-%dT00:00:00Z', "), e, Token(")"))
+			case BoolExpression:
+				return CustomExpression(Token("CASE"), e, Token("WHEN 1 THEN json('true') ELSE json('false') END"))
+			}
```
go-jet: All this types needs to be covered by tests as well. Check `TestAllTypesJSON` test for mysql.

oscar345: Okay I will take a look at the mysql tests and add them to the sqlite tests as well

go-jet:
> Should I just create a second file called json_customer_payment_sum.json with different rounding values?

That would give us two almost identical file, but we still wouldn't be sure they are truly identical.

> If I looking at the other stuff I will try to make this working but currently I am not totally sure if I can solve the problem. I will give an update about it, but maybe you can take a look at it afterwards @go-jet

It shouldn't be that complicated. Something like:

```golang
func (s *selectJsonStatement) AS(alias string) Projection {
	if statementType == SelectJsonObjStatementType {
		return Func("JSON", s.selectStatementImpl).AS(alias)
	}

	return s.selectStatementImpl.AS(alias)
}
```

tests/sqlite/main_test.go
```diff
@@ -40,9 +43,24 @@ func TestMain(m *testing.M) {
 		_, err = db.Exec(fmt.Sprintf("ATTACH DATABASE '%s' as 'chinook';", dbconfig.ChinookDBPath))
 		throw.OnError(err)
 
-		sqlSampleDB, err := sql.Open("sqlite3", dbconfig.TestSampleDBPath)
+		sql.Register("sqlite3_base64", &sqlite3.SQLiteDriver{
+			ConnectHook: func(sc *sqlite3.SQLiteConn) error {
+				return sc.RegisterFunc("BASE64_ENCODE", func(input any) any {
```
go-jet: Not sure about this. This forces every user of jet library to register `BASE64_ENCODE` function or else `SELECT_JSON` blob would not work.
I think it is better to use supported `HEX` function instead.

oscar345:
sorry (again) for the long wait, I will fix the current issues this week

oscar345:
While trying to unmarshall the json data returned from the sqlite database with the blob encoded with the hex function, I came across an issue. Since the generator creates the `[]byte` from columns with a blob type in sql, the standard unmarshal function from go does not work since it expects there to be base64 data in the json for the fields where the type is `[]byte`. So it throws an error for the data that is hex encoded.

I could see a few solutions but me none of them are ideal.

- keep using the registered base64 function 
- make blob data a string, so it would be able to unmarshal hex data into a string instead of bytes
- use a custom type for blob data in sqlite, something like `hexbytes` which will be unmarshalled and marshalled with a custom function

It could be that I am missing something here, but I think this are the only three options to make it possible.

go-jet:
Hmm, I think you were right. Using a registered base64 function is probably the cleanest. Maybe just rename `BASE64_ENCODE` to `BASE64` to follow official extension name - https://sqlite.org/base64.html.

oscar345:
Alright I will change the function name to `BASE64`, and I still have to look into the changes that need to be made for ordering (the same ones as made for mysql).

PR #538, opened 2025-12-15
Support UNIX socket connections in the MySQL generator

termermc:
The `Generate` function in `mysql_generator.go` does not support connecting to the server over a local UNIX socket, this PR adds support for that.

### DIFF
--- a/generator/mysql/mysql_generator.go
+++ b/generator/mysql/mysql_generator.go
@@ -16,8 +16,18 @@ const mysqlMaxConns = 10
 
 // DBConnection contains MySQL connection details
 type DBConnection struct {
-	Host     string
-	Port     int
+	// The server host.
+	// Has no effect if LocalSocket is set.
+	Host string
+
+	// The server port.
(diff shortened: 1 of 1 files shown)


termermc:
Are any maintainers interested in taking a look at this PR? I just remembered this while going through my email.

go-jet:
> Are any maintainers interested in taking a look at this PR? I just remembered this while going through my email.

Can this be achieved using `-dsn` or `GenerateDSN`? If so, I'd rather not extend `DBConnection` further.

termermc:
> > Are any maintainers interested in taking a look at this PR? I just remembered this while going through my email.
> 
> Can this be achieved using `-dsn` or `GenerateDSN`? If so, I'd rather not extend `DBConnection` further.

After a cursory look, it seems like GenerateDSN does support UNIX connections.

Perhaps documentation in the wiki could be updated to explain how to make a UNIX connection with GenerateDSN?

PR #548, opened 2026-01-20
Add second parameter to obj_description calls

djahandarie:
The single parameter form of obj_description is deprecated:
> obj_description ( object oid ) → text
> Returns the comment for a database object specified by its OID alone. This is deprecated since there is no guarantee that OIDs are unique across different system catalogs; therefore, the wrong comment might be returned.

https://www.postgresql.org/docs/current/functions-info.html

(It has been deprecated since at least Postgres 8.0: https://www.postgresql.org/docs/8.0/functions-info.html)

Some databases (like Materialize) only support the two-parameter form, so I'd like to merge this for compatibility there, as well as safety even in Postgres etc.


djahandarie:
(Not sure what's causing the circleci failure, but I think it's not related to this PR.)

PR #567, opened 2026-03-23
[pgx] fix(qrm): database errors swallowed as ErrNoRows on pgx v5.9.0+

gilbsgilbs:
_Note that this PR targets the `pgx` branch of go-jet._

Starting from version 5.9.0, pgx skips the `DescribePortal` round-trip for cached prepared statements[0]. When a query fails before returning any data (e.g. unique or FK constraint violation), the server never sends `RowDescription`, so `rows.FieldDescriptions()` returns nil. Then, `queryToSlicePgxV5` builds a scan context from "nil" field descriptions, gets `len == 0`, and returns `(0, nil)` without ever calling `rows.Err()`, leading callers to see `ErrNoRows` instead of the actual database error.

This commit fixes this issue by always checking for `rows.Err()` before returning.

[0] https://github.com/jackc/pgx/blob/master/CHANGELOG.md#590-march-21-2026

### DIFF
--- a/qrm/qrm_pgx_v5.go
+++ b/qrm/qrm_pgx_v5.go
@@ -132,7 +132,12 @@ func queryToSlicePgxV5(ctx context.Context, db QueryablePgxV5, query string, arg
 	if err != nil {
 		return
 	}
-	defer rows.Close()
+	defer func() {
+		rows.Close()
+		if err == nil {
+			err = rows.Err()
+		}
+	}()
 
 	scanContext, err := NewScanContextPGXv5(rows)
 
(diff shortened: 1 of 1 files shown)


go-jet:
Thanks, good catch.

qrm/qrm_pgx_v5.go
```diff
@@ -132,7 +132,12 @@ func queryToSlicePgxV5(ctx context.Context, db QueryablePgxV5, query string, arg
 	if err != nil {
 		return
 	}
-	defer rows.Close()
+	defer func() {
+		rows.Close()
+		if err == nil {
```
go-jet, line 137: The value of err should always be nil in this function. I don't think the if statement is needed.

gilbsgilbs: I'm skeptical: this is closure executes when function ends. The are three paths where `err` may not be nil as the defer fires:

1. when `NewScanContextPGXv5(...)` fails
2. when `rows.Scan(...)` fails
3. when `mapRowToSlice(...)` fails

I didn't want to swallow those errors. But maybe I'm misunderstanding something?

go-jet: Yeah, you're right, without if returned err will be overwritten.

go-jet:
LGTM. :+1:

PR #568, opened 2026-03-23
Remove pgproto3/v2 dep by upgrading pgx/v4 to v5

### LINKED ISSUE #566: Vulnerability in archived dependency github.com/jackc/pgproto3/v2
https://github.com/advisories/GHSA-jqcq-xjh3-6g23 was raised in https://github.com/jackc/pgx/issues/2507, which applies to an archived dependency of pgx v4. However, pgx is now on v5, does not have this vulnerability, and has no obvious plans to fix the vulnerability.

Is it possible to migrate onto pgx v5? I've been using it outside of jet, and believe it was more or less a drop-in replacement.

In practical terms, I'm not sure the issue is something jet users _need_ to worry about. For example, I'm running jet's generation against a postgres container in docker-compose rather than a production database that might be a target worth compromising. But it'd be handy to avoid the warning regardless.

fesiqueira:
pgx/v4/stdlib was only used in test infrastructure, but its transitive dependency on jackc/pgproto3/v2 exposed the build to GO-2026-4518 (DoS via negative field length panic in DataRow.Decode).

Replaces the pgconn.ParseConfig call via GenerateDSN with SELECT current_database(), removing the last non-test dependency on the pgx/v4.
Replaces the pgconn.ParseConfig call via GenerateDSN with url.Parse for DSN validation, and SELECT current_database() for retrieving the database name, removing the last non-test dependency on the pgx/v4.

Updates TestUInt64Overflow in tests/postgres to match pgx/v5 error format: v5 renamed Int to Int64 across pgtype to match database/sql conventions, so the overflow error now says "int64" instead of "Int8".

Bumps minimum Go version to 1.24 (required by pgx/v5).

Closes #566.


fesiqueira:
@go-jet Hi, hope you're doing well.

I see codecov/patch failed, but looking into the code I'm not sure how I'd go about testing the lines I added. I could extract the `SELECT current_database()` into something that could be replaced in the tests to force the error to happen and increase the code coverage, but I'm not sure it brings any real value. What do you think?

go-jet:
> @go-jet Hi, hope you're doing well.
> 
> I see codecov/patch failed, but looking into the code I'm not sure how I'd go about testing the lines I added. I could extract the `SELECT current_database()` into something that could be replaced in the tests to force the error to happen and increase the code coverage, but I'm not sure it brings any real value. What do you think?

I think that is enough. There are already a bunch of tests in ./tests/postgres.

LGTM. :+1:

fesiqueira:
@go-jet Hi, hope you are doing well!

Sorry for tagging you on this again, but is there any chance of getting a new release with theses changes, please? Even though it would be out of the usual release scheduled stated in the README?

joshua-temple:
Would like to also second support for an out of band release.

pgx v4 pulls in github.com/jackc/pgproto3/v2 @ 2.3.3 which is implicated by [CVE-2026-32286](https://www.cve.org/CVERecord?id=CVE-2026-32286)

go-jet:
This PR is already a part of the latest 2.15 release.

PR #574, opened 2026-03-31
Update README.md Go version to 1.24+

fesiqueira:
This is to align with current compatibility requirements after recent jackc/pgx/v5 update.


go-jet:
Thanks. :+1:

Issue #587, opened 2026-05-04
How to have uppercase field names


go-jet:
### Discussed in https://github.com/go-jet/jet/discussions/585

<div type='discussions-op-text'>

<sup>Originally posted by **navigator-dev** May  1, 2026</sup>
We have a field in our model that is an initialism (like URL, for example). Across the codebase we have other structs with this field where we use the full uppercase `URL`. From what I can tell, Jet's default behavior is to change things to Pascal case or camel case (not really sure, sorry), so that field on the model is getting set as `Url`. I have customized the generation to specifically look for that initialism and change the model field and force it to use the uppercase version, and that is working as expected. 

(a snippet of our generation script)
```go
template.Default(postgres2.Dialect).
	UseSchema(func(schema metadata.Schema) template.Schema {
		return template.DefaultSchema(schema).
			UseSQLBuilder(template.DefaultSQLBuilder().UseTable(func(table metadata.Table) template.TableSQLBuilder {
				return template.DefaultTableSQLBuilder(table).UseColumn(func(column metadata.Column) template.TableSQLBuilderColumn {
					defaultTableSQLBuilderColumn := template.DefaultTableSQLBuilderColumn(column)
					if column.Name == "url" {
						defaultTableSQLBuilderColumn.Name = "URL"
					}
					return defaultTableSQLBuilderColumn
				})
			})).
			UseModel(template.DefaultModel().
				UseTable(func(table metadata.Table) template.TableModel {
					return template.DefaultTableModel(table).
						UseField(func(c metadata.Column) template.TableModelField {
							defaultTableModelField := template.DefaultTableModelField(c)
							if c.Name == "url" {
								defaultTableModelField.Name = "URL"
							}
							return defaultTableModelField
						})
				}),
			)
	})
```

The problem we are facing is now whenever we try to use `.MODEL` or `.MODELS` we get the following error:
```
panic: missing struct field for column : url
 
-> github.com/go-jet/jet/v2/internal/jet.UnwindRowFromModel
->   /go/pkg/mod/github.com/go-jet/jet/v2@v2.13.0/internal/jet/utils.go:181

   github.com/go-jet/jet/v2/postgres.(*insertStatementImpl).MODEL
     /go/pkg/mod/github.com/go-jet/jet/v2@v2.13.0/postgres/insert_statement.go:52
```
Is there any workaround for this? Currently we're just not using the `.MODEL/S` methods, but it would simplify our code a bit.</div>

Seb-C:
[...]

PR #595, opened 2026-05-29
Expose the Snaker configuration

Seb-C:
This exposes the Snaker configuration, providing a simple way to customize the initialisms.

I made it to workaround https://github.com/go-jet/jet/issues/587, because as far as my understanding goes, alternative ways to fix it would require more important, and probably breaking changes.

The drawback is that since Snaker is used both during the code generation phase and runtime, the config needs to be set consistently in both cases.

As for the config pattern, since `snaker` is private, I opted to expose it via a global configuration in the `jet` package itself, mostly because it's the least disruptive design. However we could possibly make it public instead, that would be more consistent with what already exists via `qrm.GlobalConfig`.

What do you think?

### DIFF
--- a/internal/3rdparty/snaker/snaker.go
+++ b/internal/3rdparty/snaker/snaker.go
@@ -8,6 +8,69 @@ import (
 	"unicode"
 )
 
+// Config holds the configuration settings for the Snaker behavior.
+type Config struct {
+	// commonInitialisms, taken from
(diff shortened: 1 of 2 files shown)


go-jet:
Check: https://github.com/go-jet/jet/issues/587#issuecomment-4586710010

Seb-C:
Closing in favour of https://github.com/go-jet/jet/pull/596

PR #596, opened 2026-06-02
Added an optional struct tag to fix mapping errors

### LINKED ISSUE #587: How to have uppercase field names

Seb-C:
This adds a struct tag to manually set the column mapping in case that a non-standard name is used.

The goal of this PR is to fix https://github.com/go-jet/jet/issues/587

Non-standard names can be used with a custom config:
```go
...
.UseField(func(column metadata.Column) template.TableModelField {
	field := template.DefaultTableModelField(column)
	if column.Name == "custom_initialism" {
		field.Name = "CustomINITIALISM"
		field.Tags = append(field.Tags, fmt.Sprintf("column:%q", column.Name))
	}
	return field
})
...
```

### DIFF
--- a/generator/template/process.go
+++ b/generator/template/process.go
@@ -319,14 +319,10 @@ func processTableModels(fileTypes, modelDirPath string, tablesMetaData []metadat
 					return tableTemplate
 				},
 				"structField": func(columnMetaData metadata.Column) TableModelField {
-					field := tableTemplate.Field(columnMetaData)
-					if field.Name != DefaultTableModelField(columnMetaData).Name {
-						// If the name has been changed in UseField, we need to add the column tag
-						// to ensure that the right column will be used at runtime.
-						field.Tags = append(field.Tags, fmt.Sprintf("column:%q", columnMetaData.Name))
-					}
-
-					return field
+					return addColumnTag(
+						tableTemplate.Field(columnMetaData),
+						columnMetaData,
+					)
 				},
 				"golangComment": formatGolangComment,
 			})
--- a/generator/template/process.go
+++ b/generator/template/process.go
@@ -319,7 +319,14 @@ func processTableModels(fileTypes, modelDirPath string, tablesMetaData []metadat
 					return tableTemplate
 				},
 				"structField": func(columnMetaData metadata.Column) TableModelField {
-					return tableTemplate.Field(columnMetaData)
+					field := tableTemplate.Field(columnMetaData)
+					if field.Name != DefaultTableModelField(columnMetaData).Name {
+						// If the name has been changed in UseField, we need to add the column tag
+						// to ensure that the right column will be used at runtime.
+						field.Tags = append(field.Tags, fmt.Sprintf("column:%q", columnMetaData.Name))
+					}
+
+					return field
 				},
 				"golangComment": formatGolangComment,
 			})
--- a/internal/utils/dbidentifier/dbidentifier.go
+++ b/internal/utils/dbidentifier/dbidentifier.go
@@ -2,6 +2,7 @@ package dbidentifier
 
 import (
 	"github.com/go-jet/jet/v2/internal/3rdparty/snaker"
+	"reflect"
 	"strings"
 	"unicode"
 )
@@ -14,6 +15,28 @@ func ToGoIdentifier(databaseIdentifier string) string {
 // ToGoFileName converts database identifier to Go file name.
 func ToGoFileName(databaseIdentifier string) string {
 	return strings.ToLower(replaceInvalidChars(databaseIdentifier))
+}
+
+// GetStructFieldForColumn retrieves a column value in a struct
+func GetStructFieldForColumn(
+	structValue reflect.Value,
+	columnName string,
+) reflect.Value {
+	// Search the field using the standard name for the column
+	structField := structValue.FieldByName(ToGoIdentifier(columnName))
+	if structField.IsValid() {
+		return structField
+	}
+
+	// In case of a non-standard name, search for a matching column tag
+	for fieldIndex := range structValue.NumField() {
+		columnTag := structValue.Type().Field(fieldIndex).Tag.Get("column")
+		if columnTag == columnName {
+			return structValue.Field(fieldIndex)
+		}
+	}
+
+	panic("missing struct field for column : " + columnName)
 }
 
 func replaceInvalidChars(identifier string) string {
--- a/generator/template/model_template.go
+++ b/generator/template/model_template.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"reflect"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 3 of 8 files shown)


go-jet:
> field.Tags = append(field.Tags, fmt.Sprintf("column:%q", column.Name))

I think we can make this automatic during generation. If a field name is different from default name, add column tag.

internal/utils/dbidentifier/dbidentifier.go
```diff
@@ -16,6 +17,28 @@ func ToGoFileName(databaseIdentifier string) string {
...
+	structValue reflect.Value,
+	columnName string,
+) reflect.Value {
+	// Search the field using the standard name for the column
+	structField := structValue.FieldByName(ToGoIdentifier(columnName))
+	if structField.IsValid() {
+		return structField
+	}
+
+	// In case of a non-standard name, search for a matching column tag
```
go-jet, line 31: I guess this is fine. If there is no default match field, fall back to searching for the tag match field.

Seb-C:
@go-jet I have struggled a bit to find a good place to put the logic that adds the tag, since none of the options is ideal:
- Directly in the template: that would make it unnecessarily complicated and error-prone, it would be easier to add the logic in `TagsString`
- In the `TagsString` method of `TableModelField`: this is one option, but in terms of design it's not ideal to have a getter that magically changes the data here. Also, we don't have the column metadata here.
- In `UseField`, wrap the provided function into another one that adds this logic. Since it makes things more obfuscated and difficult to test, I decided not to go this way.
- Add the tag right before the template generation, when `tableTemplate.Field` is called. The only drawback of this option is that the new tag will only be injected in this template, and not be visible in the rest of the process.

I opted for the last option and added the commit here, let me know what you think.

go-jet:
> Add the tag right before the template generation, when tableTemplate.Field is called. The only drawback of this option is that the new tag will only be injected in this template, and not be visible in the rest of the process.

> I opted for the last option and added the commit here, let me know what you think.

I agree, it makes the most sense.

generator/template/process.go
```diff
@@ -319,7 +319,14 @@ func processTableModels(fileTypes, modelDirPath string, tablesMetaData []metadat
 					return tableTemplate
 				},
 				"structField": func(columnMetaData metadata.Column) TableModelField {
-					return tableTemplate.Field(columnMetaData)
+					field := tableTemplate.Field(columnMetaData)
+					if field.Name != DefaultTableModelField(columnMetaData).Name {
```
go-jet: maybe also expand the condition so that the tags don't already contain the `column` tag.

go-jet:
It looks, good. Maybe just one more test for the generator at: https://github.com/go-jet/jet/blob/master/tests/postgres/generator_template_test.go.

generator/template/process.go
Seb-C: Done in https://github.com/go-jet/jet/pull/596/changes/39481d655f0ffe8e6c690c884e8ce6fd94cc06a2

Seb-C:
@go-jet Thanks, I've added the condition and tests in https://github.com/go-jet/jet/pull/596/changes/39481d655f0ffe8e6c690c884e8ce6fd94cc06a2
I've isolated this logic into it's own function because it makes testing easier and I wasn't sure how to best approach testing from `generator_template_test.go`.

go-jet:
`./tests/` are integration tests. See [Readme.md](https://github.com/go-jet/jet/blob/master/tests/Readme.md) how to set them up.  

Current tests are fine, certainly better than nothing, but they still don't give guarantee that the feature actually works, and will continue to work in the future.

Seb-C:
@go-jet Ah, sorry, I did not understand that these were integration tests.

Since this PR includes both a codegen and a runtime part, the ideal way would have been to test both together IMO (but that would be more like an e2e test), but I cannot find an example in the existing tests.

Otherwise I can add two separate tests:
- One checking that the tag is present in the generated go file
- One checking that if I insert or update with a model that had the tag, the correct column name is used in the SQL query.

Would that be OK? Sorry for taking up your time, but I'm not familiar with the codebase.

go-jet:
Yeah, two tests are fine:
* one to test generator in https://github.com/go-jet/jet/blob/master/tests/postgres/generator_template_test.go 
* and the second to test the mapping - https://github.com/go-jet/jet/blob/master/tests/postgres/insert_test.go. You can use temp struct as a custom model with `column` tag.

Seb-C:
@go-jet Thanks. I added a test for the second one. For the first one, there was already an appropriate test, so I just modified the assertion to include the tag.

Seb-C:
There is an error in the CI run, but it does not seem related to my code :thinking:

go-jet:
> There is an error in the CI run, but it does not seem related to my code 🤔

Try merge with master.

Seb-C:
@go-jet The branch is already up-to-date, there is nothing to be merged. I'll try running it again, maybe it was a temporary issue with CircleCI :thinking:

Seb-C:
@go-jet Same. It's not a problem with my code but with the codecov setup, so there is nothing that I can do.
> gpg: no valid OpenPGP data found.

go-jet:
This branch is still 1 commit behind master. The issue is fixed on master.

Seb-C:
@go-jet Uh, sorry I forgot that I had to fetch from upstream :man_facepalming:

go-jet:
Integration tests are missing now.

generator/template/process.go
```diff
@@ -319,7 +319,10 @@ func processTableModels(fileTypes, modelDirPath string, tablesMetaData []metadat
 					return tableTemplate
 				},
 				"structField": func(columnMetaData metadata.Column) TableModelField {
-					return tableTemplate.Field(columnMetaData)
+					return addColumnTag(
+						tableTemplate.Field(columnMetaData),
```
go-jet, line 323: I think we can move `tableTemplate.Field(columnMetaData)` inside `addColumnTag` and get rid of this wrapper function.

Seb-C: I did it this way because it keeps a better separate of concerns between the functions and is easier to test. Also if we merged the two, it would become less obvious what each part is doing IMO.

How would you name the resulting function?

Seb-C:
@go-jet Sorry again, I messed-up my rebase while dealing with two different computers, the commit has been added now.

generator/template/process.go
go-jet: If one function parameter depends solely from another function parameter, then they are by definition the same concern.  

But it's not a big issue. It can stay this way.

go-jet:
Thanks, looks good. :+1:

Seb-C:
@go-jet Thanks for merging and for your time. Do you plan to create a release for this, or wait to aggregate more changes together?

PR #601, opened 2026-06-16
Allow a WITH (CTE) statement as an INSERT source query

### LINKED ISSUE #599: Common Table Expression with INSERT statement produces incorrect SQL for MariaDB.
**Describe the bug**
Using a CTE with an INSERT statement for MariaDB produces:

```sql
WITH cte_name AS (
    SELECT some_columns
    FROM some_table
    WHERE some_condition
)
INSERT INTO insertion_table (col_1, col_2) (
    SELECT some_col1 AS col1, some_col2 AS col2
    FROM cte_name
    WHERE another_condition
)
ON DUPLICATE KEY UPDATE col_1 = 'new_value';
```

This results in the following error: `Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'INSERT INTO...' at line XX`.

**Environment (please complete the following information):**
 - OS: windows
 - Database: MariaDB 11.8.6
-  Database driver: go-sql-driver/mysql
 - Jet version 2.15.0

**Code snippet**

```go
import (
    "database/table"
    "github.com/go-jet/jet/v2/mysql"
)

cteName := mysql.CTE("cte_name").AS(         
    mysql.SELECT(table.SomeTable.SomeColumns).
    FROM(table.SomeTable).
    WHERE(table.SomeTable.Col1.EQ(mysql.Int(1))
)

col1 := table.SomeTable.Col1.From(cteName)
col2 := table.SomeTable.Col2.From(cteName)

stmt := mysql.WITH(cteName)(
    table.InsertionTable.INSERT(
        table.InsertionTable.Col1,
        table.InsertionTable.Col2).
    QUERY(mysql.
        SELECT(
            col1, col2).
        FROM(cteName)).
    ON_DUPLICATE_KEY_UPDATE(table.InsertionTable.Col1.SET(mysql.String("new_value"))))
```

**Expected behavior**

The correct SQL should be:

```sql
INSERT INTO insertion_table (col_1, col_2)
WITH cte_name AS (
    SELECT some_columns
    FROM some_table
    WHERE some_condition
)(
    SELECT some_col1 AS col1, some_col2 AS col2
    FROM cte_name
    WHERE another_condition
)
ON DUPLICATE KEY UPDATE col_1 = 'new_value';
```

AzarAI-TOP:
## Why

`INSERT(...).QUERY(...)` only accepted a `SelectStatement`, so a CTE could not be used as the source of an INSERT. As reported in #599, MySQL and MariaDB do not support the `WITH ... INSERT ...` form, so the CTE has to be nested inside the INSERT:

```sql
INSERT INTO insertion_table (col_1, col_2)
WITH cte_name AS (
    SELECT ...
)
SELECT ... FROM cte_name
```

There was no way to express this, because `QUERY` rejected the `WITH(...)(SELECT ...)` value at compile time.

## What

- Widen `QUERY` to accept `jet.SerializerStatement` (mysql, postgres, sqlite). This is the type the underlying `ValuesQuery.Query` field already holds, and a `SELECT` still satisfies it, so existing `QUERY(SELECT(...))` calls are unaffected.
- Widen the `WITH` / `WITH_RECURSIVE` constructor return type from `Statement` to `SerializerStatement` so the result can be passed to `QUERY`. `SerializerStatement` embeds `Statement`, so this is a backward-compatible widening for existing callers.
- Fix `WITH` serialization: a CTE in a `WITH` clause must always serialize as a definition (`name AS (...)`). It previously forwarded the enclosing statement type down to the CTE, which rendered only the CTE name (its FROM-clause form) when the `WITH` was nested inside another statement, dropping the `AS (...)` body.

This follows the direction you suggested on the issue (replacing `SelectStatement` with `jet.SerializerStatement`); the `WITH` return-type widening and the serialization fix are what make that actually compile and emit correct SQL.

## Resulting SQL

```go
Link.INSERT(Link.URL, Link.Name).
    QUERY(
        WITH(
            newLinks.AS(
                SELECT(Link.URL, Link.Name).FROM(Link).WHERE(Link.ID.EQ(Int(1))),
            ),
        )(
            SELECT(Link.URL.From(newLinks), Link.Name.From(newLinks)).FROM(newLinks),
        ),
    )
```

```sql
INSERT INTO test_sample.link (url, name)
WITH new_links AS (
     SELECT link.url AS "link.url",
          link.name AS "link.name"
     FROM test_sample.link
     WHERE link.id = 1
)
SELECT new_links.`link.url` AS "link.url",
     new_links.`link.name` AS "link.name"
FROM new_links;
```

## Testing

- Added `TestInsertFromCTEQuery` in `tests/mysql/insert_test.go` (SQL assertion + executes against the DB).
- Verified the generated SQL locally and confirmed existing top-level `WITH ... SELECT/UPDATE/DELETE` output is unchanged; `go build ./...`, `go vet`, and `go test ./internal/...` pass. I don't have a local MySQL/MariaDB, so the `tests/...` integration suite runs in CI.

Fixes #599

### DIFF
--- a/internal/jet/with_statement.go
+++ b/internal/jet/with_statement.go
@@ -2,8 +2,10 @@ package jet
 
 import "fmt"
 
-// WITH function creates new with statement from list of common table expressions for specified dialect
-func WITH(dialect Dialect, recursive bool, cte ...*CommonTableExpression) func(statement Statement) Statement {
+// WITH function creates new with statement from list of common table expressions for specified dialect.
+// The returned statement is a SerializerStatement (not just a Statement) so it can be used wherever a serializable
+// statement is expected, for instance as the source of an INSERT ... QUERY(...).
+func WITH(dialect Dialect, recursive bool, cte ...*CommonTableExpression) func(statement Statement) SerializerStatement {
 	newWithImpl := &withImpl{
 		recursive: recursive,
 		ctes:      cte,
@@ -14,7 +16,7 @@ func WITH(dialect Dialect, recursive bool, cte ...*CommonTableExpression) func(s
 	}
 	newWithImpl.root = newWithImpl
 
-	return func(primaryStatement Statement) Statement {
+	return func(primaryStatement Statement) SerializerStatement {
 		serializerStatement, ok := primaryStatement.(SerializerStatement)
 		if !ok {
 			panic("jet: unsupported main WITH statement.")
@@ -44,7 +46,10 @@ func (w withImpl) serialize(statement StatementType, out *SQLBuilder, options ..
 			out.WriteString(",")
 		}
 
-		cte.serialize(statement, out, FallTrough(options)...)
+		// A CTE in a WITH clause is always serialized as a definition (name AS (...)), regardless of the enclosing
+		// statement type. Forwarding the incoming statement type instead would render the CTE in its FROM-clause form
+		// (just the name) whenever the WITH is nested, for example as the source query of an INSERT ... QUERY(WITH(...)).
+		cte.serialize(WithStatementType, out, FallTrough(options)...)
 	}
 	w.primaryStatement.serialize(statement, out, NoWrap.WithFallTrough(options)...)
 }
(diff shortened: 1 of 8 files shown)


go-jet:
Thanks. LGTM. :+1:

PR #605, opened 2026-08-13
Move test utilities out of runtime packages

MatthiasGrandl:
## Summary

- Rename internal/jet/testutils.go to internal/jet/testutils_test.go.
- Keep the default dialect used by production debug serialization in utils.go.
- Remove testing, Testify, and its dependencies from the runtime package closure.

## Why

The test utility file was compiled as production code because its name did not end in _test.go. Since it imports testing and Testify, every consumer of the SQL builder also linked Testify and gopkg.in/yaml.v3.

Before this change, go list -deps ./mysql included Testify and gopkg.in/yaml.v3. After this change, neither is present.

## Validation

- go test ./...
- golangci-lint run --timeout=30m ./...
- go list -deps ./mysql dependency check

### DIFF
--- a/internal/jet/testutils_test.go
+++ b/internal/jet/testutils_test.go
@@ -1,21 +1,10 @@
 package jet
 
 import (
-	"github.com/stretchr/testify/require"
-	"strconv"
 	"testing"
-)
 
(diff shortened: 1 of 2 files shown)


go-jet:
Thanks. LGTM :+1:

PR #610, opened 2026-08-25
Support shell wildcards in generator name filters

### LINKED ISSUE #518: Add wildcard support for `-ignore-tables` option
At the moment when we run generator on DB with existing partitions we have lots of models generated (for all the partitions), but we never need it in code. When we generate 6-12 partitions per table it's lots of things to explicitly add to `-ignore-tables`, also they are dynamic, so change names every month.
There is a task to add `-ignore-partitions` option, but it seems much complicated. 

It would be great to add wildcard support for `-ignore-tables` option, which allows to use something like:
`tableA_*` and it will ignore all partitions (depends on templare of course) like: `tableA_partX`, `tableB_partY` etc...

ChrisJr404:
Closes #518.

Right now `-ignore-tables` (and `-ignore-views`/`-ignore-enums`, plus the allow-list variants) only match names exactly. With partitioned tables the partition names change over time, so you end up listing every one by hand and editing the list whenever they rotate.

This lets the entries use shell wildcards through `path.Match`, so `payment_*` ignores `payment_2020`, `payment_2021` and so on. Names without any wildcard characters still match exactly, so existing setups keep working the same way.

Added unit tests for the matcher and the shouldSkip helpers.

### DIFF
--- a/cmd/jet/main.go
+++ b/cmd/jet/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


go-jet:
Thanks. LGTM. :+1:

PR #611, opened 2026-08-29
Add -version flag

2uent1n:
The version of the CLI is printed in the usage output, but there's no dedicated flag for it.

This adds a `-version` flag:

```
$ jet -version
v2.15.0
```


go-jet:
Looks, good, but there are some conflicts now, with recent PR merge.

PR #617, opened 2026-09-03
Improve documentation: Vitepress

nickdnk:
Hello

This is a suggestion for you. I personally think the Github wiki is terrible for documentation for a few reasons:

1. It's kind of hard to navigate and style correctly (unlike Vitepress)
2. It doesn't update with the source code, so doc changes can't ride a PR.
3. No proper search.
4. No CI checks (dead links).

This moves all documentation to vitepress and auto-deploys it on merges/pushes to master. It's a faithful 1:1 port of the existing wiki, with some language cleanups and removal of dead links.

To test it, just checkout this branch and:

```
cd docs
npm install
npm run dev
```
Future PRs (or maybe even this one) will run the doc CI and build any changed documentation for you as an artifact you can just download and open (`index.html`) directly.
 
Let me know what you think.

Edit: Note, this doc change includes the changes from https://github.com/go-jet/jet/pull/616 - I can either take it out of this PR or you can merge that PR first (assuming you want any of these). I would also suggest trimming the root README.md file significantly and pointing people to the documentation instead, so you don't have to maintain two sets of the same explanations.


safaci2000:
This is just my 2 cents, just a user of the project.

1. A key feature of docs would be something like versioned documentation. it would be nice to be able to show docs for v1 vs v2 vs v3 etc and what each one. The only one that I found that seems 
2. I'm not familiar with Vitepress but I would lean more into go based like hugo if they're feature equivalent. I had a terrible experience with Docusaurus for example that seems to generate a security alert every 3-4 days. 

Beyond that, +100 on a HTML based docs. Especially when you can make it part of the repo, write tests around it etc.

nickdnk:
> This is just my 2 cents, just a user of the project.
> 
> 1. A key feature of docs would be something like versioned documentation. it would be nice to be able to show docs for v1 vs v2 vs v3 etc and what each one. The only one that I found that seems
> 2. I'm not familiar with Vitepress but I would lean more into go based like hugo if they're feature equivalent. I had a terrible experience with Docusaurus for example that seems to generate a security alert every 3-4 days.
> 
> Beyond that, +100 on a HTML based docs. Especially when you can make it part of the repo, write tests around it etc.

For versioning, there's something like https://viteplus.github.io/versions/ - I'm sure it can be set up easily to do what we want here. It's essentially built for what you're asking for.
I'm not familiar with hugo - but vitepress is heavily used for documentation across many projects as it easily converts markdown to documentation, which was what made it possible for me to just feed the existing wiki into it. If hugo can do this just as easily, by all means go with that... but: There is nothing but markdown and 1 config file (config.mts) in the docs folder; you have to know zero JS/HTML to build what I already deployed here (please do clone and check it out, it's pretty good).

safaci2000:
There's 50 different projects that all turn your .md into html/js. 

This looks pretty clean but @go-jet would need to buy in.

PR #619, opened 2026-09-05
Add json and jsonb expression types with full separation

philon123:
Add PostgreSQL json and jsonb columns (JsonExpression, JsonbExpression). jsonb is a superset: containment, existence and deletion operators only exist on JsonbExpression, so applying them to a json expression is a compile-time error and converting requires an explicit cast (CAST(...).AS_JSONB() / AS_JSON()). This keeps the builder honest - code that compiles is valid SQL.

- JsonExpression: ->, ->>, #>, #>>, ||, EQ/NOT_EQ
- JsonbExpression: all of the above plus @>, <@, ?, ?|, ?&, -, #-, Remove/RemovePath
- ColumnJson / ColumnJsonb, JsonColumn / JsonbColumn
- JSON(value)/JSONB(value) literals, CAST(...).AS_JSON()/AS_JSONB()
- json/jsonb functions (JSON_BUILD_OBJECT/ARRAY, JSONB_BUILD_OBJECT/ARRAY, TO_JSON(B), JSON_(B)_ARRAY_LENGTH, JSON_(B)_EXTRACT_PATH, JSON_(B)_OBJECT, JSONB_PRETTY) each typed to the matching expression type
- JSON_AGG/JSONB_AGG with embedded ORDER BY (aggregate machinery)
- Generator maps json->JsonColumn, jsonb->JsonbColumn
- Cross-type tests assert the explicit-cast requirement

## SQL dialect decisions

The json/jsonb types and all operator + function wrappers are **PostgreSQL-only**. No other dialect ships a json or jsonb type in this PR (verified: `mysql/` and `sqlite/` are untouched). The generator gates on `SourceDialect`, so non-PostgreSQL databases don't regress:

- **PostgreSQL** — full support: distinct `Json`/`Jsonb` column types, the complete operator set, and all json functions below.
- **CockroachDB** — Postgres-compatible generator, but Cockroach collapses `json` into `jsonb` at the storage level. Both `json` and `jsonb` columns therefore map to a single `Json` column type here (no separate jsonb ops).
- **MySQL** — native `json` type has no jsonb distinction and no json-only operators, so it stays a plain `String` column. No `mysql.JsonExpression` is introduced.
- **SQLite** — no json column type added.

### Functions implemented (all `postgres` package)

- Builders: `JSON_BUILD_OBJECT` / `JSONB_BUILD_OBJECT`, `JSON_BUILD_ARRAY` / `JSONB_BUILD_ARRAY`, `JSON_OBJECT` / `JSONB_OBJECT`
- Coercion: `TO_JSON` / `TO_JSONB` (typed to their matching expression)
- Introspection: `JSON_ARRAY_LENGTH` / `JSONB_ARRAY_LENGTH`, `JSON_EXTRACT_PATH` / `JSONB_EXTRACT_PATH`, `JSONB_PRETTY`
- Aggregates: `JSON_AGG`, `JSONB_AGG` (and `ARRAY_AGG`) with embedded `.ORDER_BY(...)`

`json`-typed functions return `JsonExpression`; `jsonb`-typed return `JsonbExpression`; the two length functions and `JSONB_PRETTY` return scalar (`IntegerExpression` / `StringExpression`). **No json functions are implemented for MySQL or SQLite** in this PR.

## TODO / next steps

- **Reconcile with #618 (aggregate ORDER BY).** Both this PR and #618 add `internal/jet/aggregate_functions.go` + `postgres/aggregate_functions.go`. They are mergeable now but will conflict on the same files when the second merges — decide which lands first and rebase the other.
- **MySQL json type** is a natural follow-up (`mysql.JsonExpression` + json functions) but is intentionally out of scope here.
- **SQLite json1 functions** (`json_extract`, `json_each`, ...) are not covered — SQLite's json support has no jsonb subset, so it would be a distinct design rather than a port of this one.
- **json-to-row functions** (`json_to_record`, `jsonb_to_recordset`, `json_populate_record`) are not wrapped; the `->`/`#>` operators cover the common path today.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 # Test files
 gen
 .gentestdata
+.gentestdata2
 .tests/testdata/
 .gen
 .docker
--- a/generator/postgres/query_set.go
+++ b/generator/postgres/query_set.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/go-jet/jet/v2/generator/metadata"
 	postgresdialect "github.com/go-jet/jet/v2/postgres"
@@ -55,6 +56,15 @@ ORDER BY table_name;
 }
 
 func getColumnsMetaData(db *sql.DB, schemaName string, tableName string) ([]metadata.Column, error) {
+	// CockroachDB collapses the json type into jsonb, so clients can't tell a
+	// `json` from a `jsonb` column. Distinguish it so the SQL builder maps both
+	// to a single Json column type there (jsonb operators don't apply anyway).
+	var sourceDialect = postgresdialect.Dialect.Name() // "PostgreSQL"
+	var serverVersion string
+	if err := db.QueryRow("SELECT version()").Scan(&serverVersion); err == nil && strings.Contains(serverVersion, "CockroachDB") {
+		sourceDialect = "CockroachDB"
+	}
+
 	query := `
 select  
     attr.attname as "column.Name",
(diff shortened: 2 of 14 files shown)


philon123:
Hi @go-jet Please check if this PR goes in the right direction. It's a bit unfortunate that json isn't an SQL standard data type. But I do find it helpful to have type safe helpers for json / jsonb operations in postgres. The TODO section in the description above is NOT planned in this PR, unless you think it would be incomplete without.

